PR #100, opened 2026-07-03
feat(web): localize relative timestamps to the active language

### LINKED ISSUE #54: Localize relative-time date formatting for non-English languages
## Problem

The web app uses `date-fns`' `formatDistanceToNow` to render relative timestamps like "3 days ago". Called without a `locale` option, `date-fns` always renders these in English. As the app gains non-English UI languages, these relative timestamps stay English while absolute dates (formatted via `Intl.DateTimeFormat`, which follows the active language) localize — so a single view can mix languages.

## Current state

- `date-fns` `^4.4.0`.
- `formatDistanceToNow(date, { addSuffix: true })` is used at three call sites, none passing a `locale`:
  - `web/src/util/formatDate.ts` (`formatInvitedAt`)
  - `web/src/pages/SubmissionsPage.tsx` (two usages: scores-updated and last-run timestamps)
- Absolute date formatting already resolves its locale from the active UI language, so relative-time is the remaining gap.

## Why it isn't a trivial fix

`date-fns` localizes by passing a `locale` object (e.g. from `date-fns/locale/<code>`) into the format call. That requires mapping the active UI language code to a `date-fns` locale and making that locale available. The complication is that UI languages are **loaded at runtime** rather than fixed at build time, so:

- There is no static, known-ahead list of languages to import `date-fns` locales for.
- Bundling a fixed subset of locales is heavier and still won't cover an arbitrary language.
- `date-fns` locales are separate from the app's own translation strings, so a translated UI does not automatically imply a matching `date-fns` locale is present.

## Possible resolutions (to evaluate, not decided)

1. **Centralize relative-time formatting** in one helper (e.g. in `web/src/util/formatDate.ts`) and route all call sites through it, so locale handling lives in a single place.
2. **Map the active language to a `date-fns` locale and load it**, via either a bundled subset of commonly-needed locales or dynamic import; fall back to English when no matching locale is available.
3. **Use the platform `Intl.RelativeTimeFormat` API instead of `date-fns`** for relative time, since `Intl` resolves locale data from the runtime and needs no per-language bundles — at the cost of computing the unit/value to pass it (which `date-fns` currently does).

Each option has different bundle-size, accuracy, and complexity trade-offs; the right choice depends on how many languages need first-class relative-time support and whether async locale loading is acceptable.

## Acceptance

- Relative timestamps render in the active UI language where a matching locale is available.
- English remains the graceful fallback when no locale data is available for the active language.
- No regression to existing English output.

maximkern:
## Summary

`date-fns`' `formatDistanceToNow` was called without a `locale` at all three relative-time call sites, so "3 days ago" and "2 hours ago" stayed English even after the UI language was switched, while absolute dates (via `Intl`) already followed it. One view could show both at once.

The fix is a single `formatRelativeToNow` helper backed by `Intl.RelativeTimeFormat`. The three call sites now route through it.

- `formatInvitedAt` — "Invited N ago" on the Students page
- `SubmissionsPage` — "Last updated N ago" and "Last collected N ago"

`Intl` reads its locale data from the platform, which is what makes this work here, because the app's languages are sideloaded at runtime as arbitrary BCP-47 codes and there's no fixed list to bundle `date-fns` locales for. Unknown or malformed codes fall back to English. `date-fns` is gone now, since these three were its only uses.

Tested in the browser in English and with the real `pt-BR` and `ja` packs from the registry. Timestamps localize and lay out fine, and a bogus language code falls back to English.

One behavior change is worth flagging. `Intl` phrasing isn't byte-identical to `date-fns` in the sub-minute range, where it now reads "0 seconds ago" instead of the old "less than a minute ago", and it drops the "about" on hours and months. Nothing pinned those strings. I can add a flat "just now" for the sub-minute case if we'd prefer.

Closes #54

## Type of change

- [x] New feature

## Checklist

- [x] Ran `cd web && npm run check` for the web app (602 tests pass, no new lint warnings)
- [ ] Cross-binary contract — n/a, web-only
- [ ] CLI command/flag docs — n/a
- [x] Commits follow Conventional Commits

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -13,7 +13,6 @@
         "@tanstack/react-query": "^5.100.10",
         "@tanstack/react-router": "^1.168.25",
         "canvas-confetti": "^1.9.4",
-        "date-fns": "^4.4.0",
         "i18next": "^26.3.4",
         "libsodium-wrappers": "^0.8.4",
         "lucide-react": "^1.12.0",
@@ -2916,16 +2915,6 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/ljharb"
-      }
-    },
-    "node_modules/date-fns": {
-      "version": "4.4.0",
-      "resolved": "https://registry.npmjs.org/date-fns/-/date-fns-4.4.0.tgz",
-      "integrity": "sha512-+1UMbeh68lH1SegH83CGWwpb6OHHbpSgr3+s5Eww5M4CAgswBpoWS0AjTOfEJ33HiYKz1hdj/KTFprzXHmq/6w==",
-      "license": "MIT",
-      "funding": {
-        "type": "github",
-        "url": "https://github.com/sponsors/kossnocorp"
       }
     },
     "node_modules/debug": {
--- a/web/package.json
+++ b/web/package.json
@@ -22,7 +22,6 @@
     "@tanstack/react-query": "^5.100.10",
     "@tanstack/react-router": "^1.168.25",
     "canvas-confetti": "^1.9.4",
-    "date-fns": "^4.4.0",
     "i18next": "^26.3.4",
     "libsodium-wrappers": "^0.8.4",
     "lucide-react": "^1.12.0",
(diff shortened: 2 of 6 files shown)


rongxin-liu:
@maximkern Looks great, nice work!

PR #168, opened 2026-07-08
feat(web): prompt users to reload when a new version is deployed

### LINKED ISSUE #144: feat(web): prompt users to reload when a new version is deployed
## Problem

The web app is a Vite SPA served from GitHub Pages (classroom50.org). Hashed JS/CSS bundles bust themselves on each deploy (their filenames change), so the only stale-cache risk is the entry point `index.html` (and its `404.html` twin created in `.github/workflows/web-deploy.yaml`).

GitHub Pages serves HTML with a short `max-age` and an ETag, and offers **no custom-header control** — we cannot set `Cache-Control` on the document. In practice a returning user may get a stale `index.html` for a few minutes, but the genuinely bad case is a **long-lived open tab**: a user who never reloads can run an old build indefinitely, referencing old hashed bundles.

We want to mitigate this on our side, consistent with the app's 100% client-side, no-backend model.

## Proposed solution

Reuse the release stamping already present in `web/vite.config.ts` / `web/src/version.ts` (`appVersion.commit`, `version`, `buildDate`) and add a lightweight version-poll + reload prompt:

- **Emit a tiny build manifest.** Write `dist/version.json` (`{ version, commit, buildDate }`) at build time from the existing `release` object in `web/vite.config.ts`. Unhashed and tiny, so it revalidates on the same short Pages cache.
- **Poll it from the running app.** A `useVersionCheck` hook fetches `/version.json` with `{ cache: "no-store" }` on an interval (~5 min) and on `visibilitychange` / window `focus`, comparing the deployed `commit` to the compile-time `appVersion.commit`. Fail-open (any fetch/parse error = "no update"), mirroring `web/src/hooks/useSkeletonDrift.ts`. Disabled for dev/untagged builds (commit `unknown`).
- **Prompt a reload.** An `UpdateAvailableBanner` built on `web/src/components/AppBanner`, dismissible, with a button calling `window.location.reload()`. Structure modeled on `web/src/components/SkeletonDriftBanner.tsx`; mounted in `web/src/routes/_authed.tsx` alongside the existing banners.
- **i18n + tests.** New `appUpdate` copy in `web/src/locales/en.json` (all strings via `t()`); a pure `resolveUpdateAvailable({ deployedCommit, runningCommit })` verdict function unit-tested like `resolveSkeletonDrift`.

This converts "silently stale forever" into "at most one poll interval behind, then prompted", with no server-side state.

## Alternatives considered

- **Service worker / `vite-plugin-pwa`** (`registerSW` `onNeedRefresh`): can do this well but adds its own cache lifecycle and a class of "stuck on old SW" bugs. The current no-SW baseline is a fine, lower-risk starting point.
- **`<meta http-equiv="Cache-Control">` in `index.html`**: unreliable — browsers largely ignore it for the document itself.

## Additional context

Plan tracked locally in this chat session.

joshkaplan-dev:
## Summary

Prompts users to reload when a newer build of the web app has been deployed, so a long-lived open tab can't run a stale build indefinitely (GitHub Pages can't set `Cache-Control` on the entry `index.html`).

- Emit `dist/version.json` (`{ version, commit, buildDate }`) via an inline Vite plugin from the existing release object in `vite.config.ts`; a `configureServer` middleware serves the same payload in dev. No deploy-workflow changes needed — all deploy paths publish whatever `vite build` puts in `dist/`.
- New `useVersionCheck` hook polls `/version.json` with `cache: "no-store"` every 5 minutes and on focus/visibilitychange (React Query), `retry: false`, disabled for dev builds without a stamped commit. The pure `resolveUpdateAvailable()` verdict is fail-open (any fetch/parse error means no prompt) and prefix-tolerant, so a 12-char local hash vs the 40-char CI sha of the same commit never prompts.
- New dismissible `UpdateAvailableBanner` on `AppBanner` with a "Reload now" button; dismissal is per-deployed-commit, so a newer deploy after dismissing re-prompts. Mounted in `_authed.tsx` alongside the existing banners.
- New `appUpdate` strings in `en.json` (all copy through `t()`); 13 new unit tests mirroring the `resolveSkeletonDrift` style.

Closes #144

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

- [ ] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass. *(N/A — web-only change)*
- [ ] If I added or changed a CLI command or flag, I documented it in the [wiki](https://github.com/foundation50/classroom50/wiki) (not in a README). *(N/A — no CLI changes)*

### DIFF
--- a/web/src/components/UpdateAvailableBanner.tsx
+++ b/web/src/components/UpdateAvailableBanner.tsx
@@ -4,6 +4,7 @@ import { AnimatePresence } from "motion/react"
 import { useTranslation } from "react-i18next"
 
 import { AppBanner } from "@/components/AppBanner"
+import { Button } from "@/components/ui"
 import { useVersionCheck } from "@/hooks/useVersionCheck"
 
 // State the banner depends on — structural so the decision stays a pure,
@@ -50,13 +51,14 @@ export function UpdateAvailableBanner() {
           onDismiss={() => setDismissedCommit(data?.commit)}
         >
           <p className="text-base-content/70">{t("appUpdate.body")}</p>
-          <button
-            type="button"
-            className="btn btn-sm btn-success self-start"
+          <Button
+            variant="success"
+            size="sm"
+            className="self-start"
             onClick={() => window.location.reload()}
           >
             {t("appUpdate.action")}
-          </button>
+          </Button>
         </AppBanner>
       ) : null}
     </AnimatePresence>
--- a/web/src/components/UpdateAvailableBanner.tsx
+++ b/web/src/components/UpdateAvailableBanner.tsx
@@ -1,0 +1,66 @@
+import { useState } from "react"
+import { RefreshCw } from "lucide-react"
+import { AnimatePresence } from "motion/react"
+import { useTranslation } from "react-i18next"
+
+import { AppBanner } from "@/components/AppBanner"
+import { useVersionCheck } from "@/hooks/useVersionCheck"
+
+// State the banner depends on — structural so the decision stays a pure,
+// testable function (mirrors resolveDriftBannerView).
+export type UpdateBannerInput = {
+  hasUpdate: boolean
+  // Deployed commit the user dismissed. Keyed by commit so a newer deploy
+  // after dismissal re-prompts; session-only, so it also reappears on reload —
+  // fine, since reloading is exactly what fetches the new build.
+  dismissedCommit: string | undefined
+  deployedCommit: string | undefined
+}
+
+export function resolveUpdateBannerVisible(input: UpdateBannerInput): boolean {
+  const { hasUpdate, dismissedCommit, deployedCommit } = input
+  if (!hasUpdate || !deployedCommit) return false
+  return dismissedCommit !== deployedCommit
+}
+
+// Global prompt to reload when a newer build than the one this tab is running
+// has been deployed (see useVersionCheck). Reload is loop-safe: the banner
+// only shows when commits differ, and the reloaded build's commit matches. If
+// a CDN edge still serves the old index.html after reload, the banner just
+// re-shows dismissibly — fail-open.
+export function UpdateAvailableBanner() {
+  const { hasUpdate, data } = useVersionCheck()
+  const [dismissedCommit, setDismissedCommit] = useState<string>()
+  const { t } = useTranslation()
+
+  const visible = resolveUpdateBannerVisible({
+    hasUpdate,
+    dismissedCommit,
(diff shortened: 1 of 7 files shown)


rongxin-liu (dismissed):
_(Superseded — replaced by the current review below.)_

web/src/components/UpdateAvailableBanner.tsx
```diff
@@ -0,0 +1,66 @@
...
+          onDismiss={() => setDismissedCommit(data?.commit)}
+        >
+          <p className="text-base-content/70">{t("appUpdate.body")}</p>
+          <button
+            type="button"
+            className="btn btn-sm btn-success self-start"
+            onClick={() => window.location.reload()}
+          >
+            {t("appUpdate.action")}
+          </button>
```
rongxin-liu: Let's use the shared `Button` primitive here instead of the hand-rolled daisyUI classes, to match the post-#169 convention. `variant`/`size` replace the `btn-*` modifiers, and since the primitive already defaults `type="button"`, we can drop the explicit `type` and the raw `className` (keeping just the layout util):

```suggestion
          <Button
            variant="success"
            size="sm"
            className="self-start"
            onClick={() => window.location.reload()}
          >
            {t("appUpdate.action")}
          </Button>
```

Quick checklist:
- [x] Apply the suggestion above.
- [x] Add the import (see the note on the import block).
- [x] Confirm `cd web && npm run check` (tsc + eslint + prettier + vitest) still passes.

web/src/components/UpdateAvailableBanner.tsx
```diff
@@ -0,0 +1,66 @@
+import { useState } from "react"
+import { RefreshCw } from "lucide-react"
+import { AnimatePresence } from "motion/react"
+import { useTranslation } from "react-i18next"
+
+import { AppBanner } from "@/components/AppBanner"
```
rongxin-liu, line 6: Let's pull in the primitive here to go with the `Button` swap below:

```suggestion
import { AppBanner } from "@/components/AppBanner"
import { Button } from "@/components/ui"
```

rongxin-liu (changes requested):
Looks good overall! But there have been recent changes to the theme and components, so it needs some updates.

rongxin-liu (approved):

rongxin-liu (commented):
Excellent work!

PR #197, opened 2026-07-09
fix(web): harden Modal close lock and Button form-submit type

### LINKED ISSUE #170: Harden Modal/Button primitives against two latent footguns (from #169 review)
Follow-ups from the code review of #169. **Neither is a live defect** — the buggy mechanism is real and provable from the code, but no current caller reaches it. These are forward-looking hardening items so a future caller can't fall into them.

## 1. `Modal` `closeDisabled` doesn't guard the programmatic close path

`closeDisabled` is the "don't let the user dismiss this modal while a submit is in flight" lock. It's enforced on Esc (`onCancel` veto), the close-X button, and the backdrop button — but **not** on the controlled-`open` effect:

```ts
// web/src/components/ui/Modal.tsx
useEffect(() => {
  if (open === undefined) return
  const dialog = dialogRef?.current ?? internalRef.current
  if (!dialog) return
  if (open && !dialog.open) dialog.showModal()
  if (!open && dialog.open) dialog.close()   // <-- ignores closeDisabled
}, [open, dialogRef])
```

**Mechanism:** a parent rendering `<Modal open={x} closeDisabled={submitting}>` that flips `open` to `false` while `submitting` is `true` closes a modal the guard was protecting. The lock covers *human* dismissal but not the parent's own state transition.

**Why latent:** every current `closeDisabled` caller either guards its own close (e.g. `AddStudent`'s `closeDialog` does `if (submitting) return`) or is ref-driven with no `open` prop (`ReuseModalShell`, etc.), so the effect early-returns. The hand-rolled `ConfirmModal` this primitive replaced funneled all closes through an `isSubmitting` check; the primitive dropped that for the programmatic path.

**Suggested fix:** either
- `if (!open && dialog.open && !closeDisabled) dialog.close()` (add `closeDisabled` to the effect deps), or
- explicitly document that `closeDisabled` only blocks *user* dismissal and callers must not set `open=false` while submitting.

Add a test: `<Modal open closeDisabled>` then flip `open=false` and assert the dialog stays open (or document + test the intended contract).

## 2. `Button` `type ?? "button"` can silently strip implicit form-submit

```tsx
// web/src/components/ui/Button.tsx
<button type={type ?? "button"} ...>
```

**Mechanism:** a bare HTML `<button>` inside a `<form>` defaults to `type="submit"`; the `Button` primitive defaults to `type="button"`. A future migration that replaces a bare implicit-submit `<button>` with `<Button>` and forgets `type="submit"` would silently stop that form from submitting — invisible in a "visual" diff.

**Why latent:** audited in #169 — every migrated `<Button>` submit site already carries an explicit `type=\"submit\"`, and the forms use TanStack Form `onSubmit` + `e.preventDefault()`, so none rely on implicit button-type submission. No form is broken today.

**Suggested fix:** add a lint/grep guard that flags a `<Button>` descendant of a `<form>` lacking an explicit `type`, plus a render test asserting Enter/click still fires `onSubmit`. (Documenting the `type="button"` default on `ButtonProps` would also help.)

---

Discovered during the multi-agent review of #169 (theme redesign + shared UI primitive standardization).

rongxin-liu:
## Summary

Closes #170 — two forward-looking hardening items from the #169 review. Neither was a live defect, but both mechanisms were provable from the code and unguarded.

- **Modal `closeDisabled` now guards the programmatic close path.** The open-sync effect previously called `dialog.close()` unconditionally, so a parent flipping `open=false` while a submit was in flight would dismiss a modal the lock was protecting. It now checks `!closeDisabled` (added to the effect deps); the pending close applies once the lock releases. Doc comment updated to reflect the wider contract.
- **`Button` form-submit type footgun guarded.** A bare `<button>` in a `<form>` defaults to `type="submit"`; the `Button` primitive defaults to `type="button"`. Added a `no-restricted-syntax` eslint rule (repo's existing `warn` convention) that flags any `<Button>` descendant of a `<form>` lacking an explicit `type`, plus render tests asserting click/Enter submit behavior. Documented the default on `ButtonProps`.
- **Fixed three latent sites the new rule caught** — cancel/close buttons inside forms in `AddStudent`, `EditStudentForm`, and `CreateAssignmentForm` that relied on the implicit default. Made their `type="button"` explicit (no behavior change; the intent is now declared and robust).

## Test plan

- [x] `tsc -b` — clean
- [x] `eslint .` — 0 errors, 0 remaining form-button warnings
- [x] `prettier --check` — clean on all touched files
- [x] `vitest run` — 1047 passed (95 files), including 4 new tests:
  - Modal holds open against `open=false` while `closeDisabled`, then closes when the lock releases
  - Button `type="submit"` fires form `onSubmit` on click
  - Button `type="submit"` fires form `onSubmit` on Enter
  - Button default type does not submit

### DIFF
--- a/web/eslint.config.js
+++ b/web/eslint.config.js
@@ -93,14 +93,15 @@ export default defineConfig([
             'Prefer the accessible <Spinner> component over a bare `loading loading-spinner` span (it adds role="status" + an sr-only label). In-button spinners may stay inline if the button already has an accessible name.',
         },
         {
-          // A bare <button> inside a <form> defaults to type="submit"; our
-          // <Button> primitive defaults to type="button". A <Button> under a
-          // <form> without an explicit `type` is almost certainly a silently
-          // broken submit (or a click-handler button that should say
-          // type="button" to make the intent explicit). Descendant selector, so
-          // it catches nesting at any depth.
+          // Warn when a <Button> sits in a form without an explicit `type`: a
+          // bare <button> defaults to submit, our <Button> defaults to "button",
+          // so an unmarked one is a likely silent no-op submit. Advisory only —
+          // the Button default + form-submit tests are the real enforcement.
+          // Two deliberate limits: matches <form> AND `<Card as="form">` (the
+          // app uses the latter); same-file lexical only, so a <Button> in a
+          // child component rendered inside a form isn't reachable here.
           selector:
-            "JSXElement[openingElement.name.name='form'] JSXElement[openingElement.name.name='Button']:not(:has(JSXAttribute[name.name='type']))",
+            ":matches(JSXElement[openingElement.name.name='form'], JSXElement:has(JSXAttribute[name.name='as'][value.value='form'])) JSXOpeningElement[name.name='Button']:not(:has(JSXAttribute[name.name=/^(type|as|href)$/]))",
           message:
             'A <Button> inside a <form> needs an explicit `type`: add type="submit" for the submit action or type="button" for a click handler. The <Button> default is "button", which silently disables implicit form submit.',
         },
--- a/web/eslint.config.js
+++ b/web/eslint.config.js
@@ -92,6 +92,18 @@ export default defineConfig([
           message:
             'Prefer the accessible <Spinner> component over a bare `loading loading-spinner` span (it adds role="status" + an sr-only label). In-button spinners may stay inline if the button already has an accessible name.',
         },
+        {
+          // A bare <button> inside a <form> defaults to type="submit"; our
+          // <Button> primitive defaults to type="button". A <Button> under a
+          // <form> without an explicit `type` is almost certainly a silently
+          // broken submit (or a click-handler button that should say
+          // type="button" to make the intent explicit). Descendant selector, so
+          // it catches nesting at any depth.
+          selector:
+            "JSXElement[openingElement.name.name='form'] JSXElement[openingElement.name.name='Button']:not(:has(JSXAttribute[name.name='type']))",
+          message:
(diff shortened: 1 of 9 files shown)


rongxin-liu:
Follow-up commit `ef6f9bc` addresses a multi-agent code review of this PR (correctness / testing / maintainability / project-standards / frontend-races / adversarial) plus a simplification pass.

**Guard fidelity (the eslint rule was weaker/narrower than it looked):**
- The selector now matches `<Card as="form">` in addition to a literal `<form>` — the app's real forms use the former, so the original rule was blind to them. Surfaced and fixed one genuinely unguarded in-form button (`CreateClassroomForm`'s regenerate button now has explicit `type="button"`).
- Scoped the `type` check to the Button's own opening element, so a nested typed child (e.g. `<Button><input type=…></Button>`) no longer suppresses the warning.
- Excluded anchor-mode Buttons (`as="a"`/`href`) — an `<a>` has no submit semantics.
- Each selector branch was verified against the real `@typescript-eslint/parser` AST via esquery.

**Comment accuracy:** the `ButtonProps.type` doc comment cited an `eslint-plugin-react button-has-type` rule that isn't enabled anywhere; repointed it at the actual `no-restricted-syntax` guard and trimmed duplicated rationale across `Button`/`Modal`/`eslint.config.js` per the repo's say-it-once convention.

Note: the rule stays `warn` (advisory), consistent with the repo's existing warning convention — the Button `type="button"` default + the form-submit tests are the real enforcement.

Verification: `tsc -b`, `eslint .` (0 errors, 0 new-rule warnings across the app), `prettier --check`, and the full vitest suite (1047 passing) all green.

Known follow-ups (not in this PR): a `RuleTester` meta-test to pin the selector against future rot, and Modal reopen-while-locked test coverage.

Issue #198, opened 2026-07-09
test(web): pin the Button-in-form lint guard with a RuleTester meta-test


rongxin-liu:
Follow-up from the multi-agent review of #197 (which hardened the `<Button>`-in-`<form>` `no-restricted-syntax` guard added for #170). **Not a live defect** — the guard currently behaves correctly and was verified empirically during the review. This is forward-looking hardening so the guard can't silently rot, plus two small test gaps.

## Background

#197 added a hand-written esquery selector in `web/eslint.config.js` that warns when a `<Button>` inside a `<form>` (or `<Card as="form">`) lacks an explicit `type` — because our `Button` primitive defaults to `type="button"` while a bare `<button>` defaults to `type="submit"`, so an unmarked in-form button is a likely silent no-op submit.

The selector is intricate and its correctness is entirely implicit:

```js
":matches(JSXElement[openingElement.name.name='form'], JSXElement:has(JSXAttribute[name.name='as'][value.value='form'])) JSXOpeningElement[name.name='Button']:not(:has(JSXAttribute[name.name=/^(type|as|href)$/]))"
```

During review this selector went through **three** correctness bugs before landing (a descendant `:has(type)` that let a nested child's `type` suppress the warning; a `<Card as="form">` blind spot; over-matching `as="a"` anchors). Each was caught only by hand-probing the AST. Nothing in the repo pins the final behavior.

## 1. Pin the selector with a RuleTester meta-test (main ask)

**What we're hardening:** the guard is a *silent-pass verification mechanism* — if a future edit (an eslint/esquery upgrade, a selector tweak, a Button/Card rename) makes the selector match nothing, it turns into a permanent green PASS with no signal. There is no `RuleTester` anywhere in `web/` today.

**Fix:** add a fixture-based test asserting the selector both fires and stays quiet on the known shapes. It should cover, at minimum:

- **Should warn:** `<form><Button>x</Button></form>`; `<Card as=\"form\"><Button>x</Button></Card>`; a `<Button>` nested deeper inside a form (e.g. wrapped in a `<div>`/`<span>`); a typeless `<Button>` that *wraps a typed child* like `<Button><input type=\"text\"/></Button>` (the descendant-`:has` regression that shipped once).
[...]

PR #247, opened 2026-07-14
test(web): pin <Button> form lint guard and strengthen Modal close-lock coverage

AkshajSinghal:
## What's added:
- Add noRestrictedSyntax.test.ts to pin the `no-restricted-syntax` selector and verify known warn/no-warn cases
- Extend Modal.test.tsx to cover reopen-while-locked and deferred `onClose` behavior

This adds a regression guard so the hand-written ESLint selector cannot silently stop matching real problematic JSX, while also tightening the modal state machine behavior in tests.

Closes Issue #198 

## Validation:
 - `npm run test` ran successfully with no test failures, except for some formatting warnings, which is unrelated to our logic and have been safely ignored
 - `npx vitest run` passed all 126 files, 0 errors


AkshajSinghal:
image

web/src/eslint/noRestrictedSyntax.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    expect(buttonRule).toBeDefined()
+    expect(buttonRule?.selector).toBe(BUTTON_FORM_SELECTOR)
+  })
+
+  it("warns for a typeless <Button> inside a <form>", async () => {
+    const source = `
+      import { Button } from "./Button"
+      export function App() {
+        return <form><Button>Go</Button></form>
```
rongxin-liu, line 47: Every warn/no-warn fixture uses a **zero-attribute** `<Button>`, so the selector's core exclusion clause — `:not(:has(JSXAttribute[name.name=/^(type|as|href)$/]))` — is never actually exercised against a Button that carries a *non-safe* attribute. That leaves a green-while-red hole: weakening the clause to `:not(:has(JSXAttribute))` keeps all six cases passing while the rule silently stops flagging the most common real offender, `<form><Button onClick={save}>Save</Button></form>`.

Since the whole point here is a guard that can't silently stop matching, we should pin that clause directly. Worth adding:
- a **warn** case for a typeless Button with a non-safe attr: `<form><Button onClick={save}>Save</Button></form>` → expect `1`
- a symmetric **safe** case where a typed Button also carries extras: `<Button type="submit" onClick={h} className="x">Go</Button>` → expect `0`

web/src/eslint/noRestrictedSyntax.test.ts
```diff
@@ -0,0 +1,112 @@
...
+          <form>
+            <Button><input type="text" /></Button>
+          </form>
+        )
+      }
+    `
+    expect(await lintMessageCount(source)).toBe(1)
+  })
+
+  it.each([
```
rongxin-liu, line 148: The only case covering the `as` exemption also carries `href`, so dropping `as` from the `/^(type|as|href)$/` alternation would regress silently. Maybe add a no-warn case that isolates `as` alone inside a form:

```tsx
`<form><Button as="div">Go</Button></form>` // expect 0
```

This and the look-alike-attribute case fall out naturally once we're exercising the exclusion clause (see the comment above).

web/src/eslint/noRestrictedSyntax.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      .filter((entry): entry is Record<string, unknown> => typeof entry === "object" && entry !== null)
+    const buttonRule = ruleObjects.find(
+      (entry) => entry.message === BUTTON_FORM_MESSAGE,
+    )
+
+    expect(buttonRule).toBeDefined()
+    expect(buttonRule?.selector).toBe(BUTTON_FORM_SELECTOR)
+  })
+
+  it("warns for a typeless <Button> inside a <form>", async () => {
```
rongxin-liu, line 43: The warn/no-warn helper filters on an exact `message` string that's independent of the selector, so a selector that broadens or narrows its match set while keeping the same message passes silently. One narrow example not yet covered: dropping the `^…$` anchors on the exemption regex would then exempt look-alike attrs. Worth a warn case to pin it:

```tsx
`<form><Button data-type="x">Go</Button></form>` // expect 1
```

web/src/eslint/noRestrictedSyntax.test.ts
```diff
@@ -0,0 +1,112 @@
...
+import { fileURLToPath } from "url"
+import { describe, expect, it } from "vitest"
+import { ESLint } from "eslint"
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
+const configPath = path.resolve(__dirname, "../../eslint.config.js")
+const projectRoot = path.resolve(__dirname, "../..")
+
+const BUTTON_FORM_SELECTOR =
```
rongxin-liu: `BUTTON_FORM_SELECTOR` / `BUTTON_FORM_MESSAGE` are verbatim copies of the strings inlined in `eslint.config.js` with no compile-time link — the "pins" test only catches a *one-sided* edit, and a coordinated edit to both strings restores green. 

Let's remove the duplication rather than pin it: extract both to a small shared module (e.g. `web/src/eslint/buttonFormRule.ts` exporting `buttonFormSelector` + `buttonFormMessage`), import them into the config's rule and into this test. Drift then becomes impossible by construction, and the fragile config-introspection block (`~92–108`) can be dropped.

rongxin-liu (changes requested):

rongxin-liu (commented):
Thanks for the PR! Left a couple comments worth taking a look.

AkshajSinghal:
Sure, fixing the issues!

AkshajSinghal:
All the changes done:
- Move selector/message to buttonFormRule.js and import from
  eslint.config.js and noRestrictedSyntax.test.ts
- Remove brittle config-introspection assertion from the test
- Add tests covering:
  - typeless `<Button onClick={...}>` inside `<form>` (warn)
  - `<form><Button as="div">` (no warn)
  - look-alike attribute (`data-type`) inside `<form>` (warn)
  - typed `<Button type="submit" onClick=... className=...>` (no warn)
- Add root package.json with test script delegating to web so
  `npm run test` works from repo root
- All tests run locally (Vitest): 126 files, 1461 tests passing.

What do you think?

web/src/eslint/noRestrictedSyntax.test.ts
AkshajSinghal: Resolved in the commit [496ae0f](https://github.com/foundation50/classroom50/pull/247/commits/496ae0f9d31a5b24795713247d1db897bdf02a9b)

web/src/eslint/noRestrictedSyntax.test.ts
AkshajSinghal: Resolved in the commit [496ae0f](https://github.com/foundation50/classroom50/pull/247/commits/496ae0f9d31a5b24795713247d1db897bdf02a9b)

web/src/eslint/noRestrictedSyntax.test.ts
AkshajSinghal: Resolved in the commit [496ae0f](https://github.com/foundation50/classroom50/pull/247/commits/496ae0f9d31a5b24795713247d1db897bdf02a9b)

web/src/eslint/noRestrictedSyntax.test.ts
AkshajSinghal: Resolved in the commit [496ae0f](https://github.com/foundation50/classroom50/pull/247/commits/496ae0f9d31a5b24795713247d1db897bdf02a9b)

package.json
```diff
@@ -0,0 +1,7 @@
+{
```
rongxin-liu: Let's drop this root `package.json` from the PR entirely, it's unrelated to the eslint-guard/Modal work and pulls the change out of scope.

This repo is a monorepo of independently-built parts (`web/`, `cli/*`, `schemas/`) with no existing root manifest and path-scoped release-please tracks, so a top-level `classroom50-root` manifest implies a JS-workspace root that doesn't exist.

web/src/eslint/buttonFormRule.ts
```diff
@@ -0,0 +1,5 @@
+export const buttonFormSelector =
```
rongxin-liu, line 1: The new coverage closed the `onClick` hole, nice.

There's still a sibling gap on the same exclusion clause: every warn fixture uses `onClick`/`data-type` as the extra attr, so none pins an ordinary `<Button>` prop. Widening the allowlist to e.g. `/^(type|as|href|variant)$/` passes all current fixtures while silently un-linting `<form><Button variant="primary">Go</Button></form>` — our most common real Button shape.

We should pin the semantics from the warn side. Maybe add to `noRestrictedSyntax.test.ts`:
- `<form><Button variant="primary">Go</Button></form>` // expect 1
- `<form><Button className="x">Go</Button></form>` // expect 1

That way any broadening of the safe-attr list flips a test red.

web/src/eslint/noRestrictedSyntax.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    const source = `
+      import { Button } from "./Button"
+      export function App() {
+        return <form><Button data-type="x">Go</Button></form>
+      }
+    `
+    expect(await lintMessageCount(source)).toBe(1)
+  })
+
+  it.each([
```
rongxin-liu, line 148: The `as` exclusion is now pinned in-form (`<form><Button as="div">`), but the `type` and `href` arms of `/^(type|as|href)$/` are only exercised inside a `<div>` here, so they're spared by the *ancestor* clause, not the *exclusion*. Dropping `type` (the rule's primary safe path) or `href` from the alternation would keep this whole suite green.

Let's move the two safe arms in-form so the exclusion itself is under test:
- `<form><Button type="submit">Go</Button></form>` // expect 0
- `<form><Button href="/">Go</Button></form>` // expect 0

web/src/eslint/noRestrictedSyntax.test.ts
```diff
@@ -0,0 +1,120 @@
+// @vitest-environment node
+import path from "path"
+import { fileURLToPath } from "url"
+import { describe, expect, it } from "vitest"
+import { ESLint } from "eslint"
+import { buttonFormSelector, buttonFormMessage } from "./buttonFormRule.js"
```
rongxin-liu: One thing worth being aware of: now that both the config and this test import the same `buttonFormSelector`/`buttonFormMessage`, and the helper filters lint output by that same imported message, the two agree by construction, so the behavioral fixtures are now the *only* guard against a wrong selector. That's fine once the coverage above (variant + in-form type/href) is in.

Optionally, a snapshot or independent hand-written literal of `buttonFormSelector` would restore an explicit drift check on top.

rongxin-liu (changes requested):

AkshajSinghal:
Thanks for the detailed feedback, I've added explicit tests for variant/className warnings and form-scoped type/href exemptions, and removed the temporary root package.json change so the monorepo layout remains intact. All tests are passing!

package.json
AkshajSinghal: Resolved with commit [60f703c](https://github.com/foundation50/classroom50/pull/247/commits/60f703c02c6fccb8eed79db81e327f0b4754a2dd)

web/src/eslint/buttonFormRule.ts
AkshajSinghal: Resolved with commit [60f703c](https://github.com/foundation50/classroom50/pull/247/commits/60f703c02c6fccb8eed79db81e327f0b4754a2dd)

web/src/eslint/noRestrictedSyntax.test.ts
AkshajSinghal: Resolved with commit [60f703c](https://github.com/foundation50/classroom50/pull/247/commits/60f703c02c6fccb8eed79db81e327f0b4754a2dd)

web/src/eslint/noRestrictedSyntax.test.ts
AkshajSinghal: Resolved with commit [60f703c](https://github.com/foundation50/classroom50/pull/247/commits/60f703c02c6fccb8eed79db81e327f0b4754a2dd)

rongxin-liu (dismissed):
Thank you!

rongxin-liu (approved):

PR #340, opened 2026-07-20
feat(web): add RTL language support (Arabic, Hebrew, Farsi, Urdu)

### LINKED ISSUE #141: feat(web): support Arabic (RTL layout + ar language pack)
## Goal

Support **Arabic (`ar`)** as a first-class UI language. Arabic is one of the largest uncovered populations (~335M+ speakers, L1+L2) and is handled reasonably well by frontier LLMs, so our Bedrock translation pipeline can produce a usable pack. The real work is **right-to-left (RTL) layout**, which the app has no support for today.

Rendering Arabic strings in the current LTR layout (sidebar on the wrong side, mirrored spacing, back-arrows pointing the wrong way) reads as broken to an Arabic speaker, so shipping the pack alone is not enough.

### 1. Arabic language pack (small)
- [ ] Add `"ar"` to `web/src/locales/targets.json` so CI (`translate-locales.yaml` + Bedrock) generates and publishes it. Use a strong Bedrock model — Arabic is high-resource but harder than the European targets.
- `ar` already passes `langCodeSchema` (BCP-47), so it flows through the loader, registry, and `verify_locale.py` with no code changes.
- `formatDate.ts` already localizes dates/relative times off `i18n.language` via `Intl.*`, so Arabic dates/numerals work once `ar` is active — no change expected.

### 2. RTL layout (the actual project)
Nothing sets text direction anywhere today: the only thing written to `<html>` is `data-theme` (see `useTheme.ts` + the anti-flash script in `index.html`), and `<html lang="en">` is hardcoded.

- [ ] **Drive `dir`/`lang` off the active language.** Add an `RTL_LANGS` set (`ar`, `he`, `fa`, `ur`, …) and on `languageChanged` set `document.documentElement.dir` and `.lang`. Mirror the existing theme pattern (a small effect, or in `web/src/i18n/index.ts` where language activation already runs).
- [ ] **Update the anti-flash inline script in `index.html`** to read `classroom50:lang` and set `dir` before React mounts, so first paint isn't LTR-then-flip.
- [ ] **Convert physical Tailwind classes to logical ones** (~50–60 occurrences across ~35 files; these do NOT auto-flip):
  - `ml-*/mr-*` → `ms-*/me-*`
  - `pl-*/pr-*` → `ps-*/pe-*`
  - `text-left/right` → `text-start/end`
  - `left-*/right-*` → `start-*/end-*`
  - `border-l/r` → `border-s/e`, `rounded-l/r` → `rounded-s/e`
  - Biggest offender: the sidebar drawer (`components/drawer/index.tsx`) — hard `left-*`/`right-*` positioning and a `border-l-2` active indicator. Also modal close buttons (`absolute right-3 top-3` → `end-3`) in several files.
- [ ] **Flip directional icons** (~15 files use `ChevronLeft/Right`, `ArrowLeft/Right` for next/back/expand — wizard nav in `OrgSetupPage`, drawer collapse, disclosure carets). A "back" arrow must point the other way in RTL. Leave pure vertical `rotate-90/180` carets alone.
- [ ] **Keep code/identifiers LTR** inside RTL text — repo names, URLs, org slugs, and `<pre>` CLI blocks (e.g. `AcceptAssignmentPage`) should be wrapped `dir="ltr"` / `unicode-bidi: isolate` so they don't scramble.

### 3. Guardrail
- [ ] Add a lint/audit check flagging new physical directional classes (`ml-/mr-/pl-/pr-/text-left/right`, etc.) so LTR-only classes don't creep back after the migration.

## Notes
- Stack is Tailwind v4 + DaisyUI v5. DaisyUI v5 is direction-aware and flips its own components (drawer, dropdowns, tabs, toggles) automatically once `dir="rtl"` is on the root, so a lot mirrors for free.
- Low architectural risk; the bulk is mechanical class conversion + a small direction hook.

## Suggested sequencing
1. `dir`/`lang` activation (hook + anti-flash script) — gives a real RTL canvas.
2. Add `ar` to `targets.json` (or hand-load a pack) for Arabic text to eyeball.
3. Sweep physical→logical classes (drawer, modals, wizards first).
4. Flip nav icons; isolate code/identifier spans as LTR.
5. Visual QA pass in Arabic + lint guard.

joshkaplan-dev:
## Summary

Adds full right-to-left language support to the web app and enables Hebrew, Arabic, Farsi, and Urdu in the translation pipeline.

- **Direction wiring** — new `web/src/i18n/direction.ts` mirrors `<html lang>`/`<html dir>` to the active language (BCP-47 primary-subtagmatch, so sideloaded regional codes like `ar-EG` work). An anti-flash
  script in `index.html` applies the stored language's direction before
  first paint, pinned to `RTL_LANGS` by a parity test. `ar`/`he`/`fa`/`ur`
  added to `targets.json` — packs generate on the first post-merge
  translation run.
- **Logical CSS** — all ~66 physical directional Tailwind classes converted to logical equivalents (`ms-`/`me-`/`ps-`/`pe-`/`start-`/`end-`/ `text-start`/`border-s`) across 40 files so the layout mirrors automatically when the condition is set to `dir="rtl"`.
- **Icon mirroring** — shared `rtlFlip` (`rtl:-scale-x-100`) on semantically directional icons only (back arrows, drill-in/collapse chevrons); hover nudges use exclusive `ltr:`/`rtl:` variants; `‹` glyphs in nav strings replaced with real mirrored icons.
- **Sentence-fragment refactor** — all ~150 split `_prefix`/`_suffix`/`_1..n` i18n keys merged into single-sentence keys via `<Trans>` with named tags (`<repo>{{repo}}</repo>`) or plain interpolation, so translators control word order. New `MonoLtr` primitive LTR-isolates identifiers (repo names,`@usernames`, URLs, commands) inside RTL sentences.
- **Pipeline + guardrails** — `verify_locale.py` validates `<tag>` markers and warns on missing CLDR plural categories (i18next renders *English* for counts whose category key is absent — Arabic needs zero/one/two/few/many/other); `TRANSLATION_PROMPT.md` rewritten for the tag contract; pack ingest rejects attribute-bearing markup tags (a
  hostile sideloaded pack could otherwise repoint trusted link hrefs via`<Trans>` attribute merging); ESLint rule + `audit_i18n.py` backstop fail CI on new physical directional classes or returning split keys.

**Testing:** 1,783 vitest + 64 pytest, lint, typecheck, prettier, and`audit_i18n.py --strict` all green. Manual QA with a pseudo-hebrew pack through the real install path: no-flash RTL reload, mirrored sidebar/tooltips/chevrons, LTR-isolated identifiers, `ar-EG` subtag detection, `?lang=en` round-trip.

**Post-merge follow-ups:** confirm the translate workflow generates the four new packs, then spot-check a real pack in the deployed app via `?lang=ar`(including the two `→` glyphs in `preflightRoleChangeDetail` and the org-settings prose).

Closes #141

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

- [x] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass. *(N/A — no schema changes; the i18n key changes are web-internal)*
- [x] If I added or changed a CLI command or flag, I documented it in the [wiki](https://github.com/foundation50/classroom50/wiki) (not in a README). *(N/A — no CLI changes)*


web/src/i18n/index.ts
```diff
@@ -28,6 +29,17 @@ void i18n.use(initReactI18next).init({
...
 
+// Keep <html dir>/<html lang> in step with the active language. Registered
+// before the changeLanguage chain below so every switch — startup activation,
+// deep links, and user selections — updates document direction. The explicit
+// call covers the case where the stored language is the init default ("en")
+// and no languageChanged event fires. Skipped without a DOM (node tests import
+// this module).
+if (typeof document !== "undefined") {
+  i18n.on("languageChanged", applyDocumentDirection)
+  applyDocumentDirection(i18n.language)
```
rongxin-liu: This runs at module eval with `i18n.language === "en"` (the init default), so it resets `dir=ltr`/`lang=en` and undoes the index.html anti-flash script. The common case survives on microtask-before-paint timing, but there's an edge: line 57 only calls `changeLanguage(stored)` when the stored language's pack is installed — so if a user has a persisted RTL `lang` but the pack isn't installed, this `ltr` reset is never undone and the RTL direction is permanently lost for that load.

Maybe seed the direction from the stored language instead? Perhaps `i18n.init({ lng: getStoredLang(), … })` so `i18n.language` already matches before this line, or drop this unconditional call and let the async chain be the only writer. Worth a quick RTL-persisted-but-uninstalled test either way.

web/src/eslint/directionalClassRule.ts
```diff
@@ -0,0 +1,28 @@
...
+// is matched at start-of-string, after whitespace, or after a variant colon,
+// with an optional negative prefix. Deliberately NOT matched: translate-x
+// (no logical form — pair with ltr:/rtl: variants), space-x/inset-x (already
+// logical in Tailwind v4), DaisyUI tooltip-left/right (no logical variant —
+// pair with rtl: overrides), and grid col-start/justify-start.
+//
+// Kept in sync with PHYSICAL_CLASS_RE in web/src/locales/audit_i18n.py — the
+// line-based CI backstop that also sees template-literal chunks in .ts class
+// recipes, which these AST selectors cannot reach.
+export const directionalClassPattern =
```
rongxin-liu, line 13: This pattern and `PHYSICAL_CLASS_RE` in audit_i18n.py are bound only by the "keep in sync" comments, each with its own fixtures — and the Python copy already carries a quote/backtick opening the test notes "a blind re-sync would drop." A stale Python regex is the risky direction since it's the backstop for `.ts` class recipes the AST rule can't see.

Could we add a small parity test that derives/compares the two token sets (or extract one shared token list both consume)? Would make the sync a tested contract rather than a hand-sync.

web/src/eslint/directionalClassRule.ts
```diff
@@ -0,0 +1,28 @@
...
+  "(?:^|[\\s:])-?(?:(?:scroll-)?[mp][lr]|left|right)-(?:\\d|\\[|auto|full|px)" +
+  "|(?:^|[\\s:])text-(?:left|right)(?![A-Za-z0-9_-])" +
+  "|(?:^|[\\s:])(?:border|rounded)-(?:[lr]|t[lr]|b[lr])(?![A-Za-z])"
```
rongxin-liu: Minor/latent: `float-left`/`float-right` and `clear-left`/`clear-right` are real Tailwind physical utilities (with logical `float-start`/`float-end`, `clear-start`/`clear-end` forms) that neither this regex nor the Python one matches. None exist in source today, so no active break — but worth adding here + in PHYSICAL_CLASS_RE so a future `float-left` can't ship green and silently break RTL.

web/src/locales/audit_i18n.py
```diff
@@ -91,6 +112,27 @@
 # literals that look like code/config, not prose worth translating
 CODEISH_RE = re.compile(r"^[a-z0-9]+(?:[-_/][a-z0-9]+)*$")  # ubuntu-latest, foo_bar
 
+# The retired sentence-fragment suffixes (SPLIT check). Match on the literal
+# underscore suffix so camelCase keys like nav.allClassesLink stay legal.
+# _<digits> covers the numbered variants (_1, _2, ...).
+SPLIT_SUFFIX_RE = re.compile(r"_(?:prefix|suffix|middle|from|emphasis|link|\d+)$")
```
rongxin-liu: The SPLIT gate is underscore-only, and there are a handful of camelCase fragment keys still living in en.json — `orgMembers.idSuffix`, `students.githubIdSuffix`, `assignments.form.passThresholdSuffix`, `published.servedUnlistedSuffix`, `classes.toolbar.termPrefix`/`sortPrefix` — that are exactly the retired concatenated-fragment convention this gate blocks, but they evade it (the docstring even flags this).

Should we either fold those into single `{{placeholder}}` sentences, or extend the smell check to warn on camelCase `Prefix`/`Suffix`/`Middle` tails? Otherwise the green here is a bit misleading.

web/src/pages/assignments/AssignmentsTable.tsx
```diff
@@ -62,17 +62,21 @@ const DeleteAssignmentButton = ({
...
-          </>
+          <Trans
+            i18nKey="assignments.table.deleteDescription"
+            values={{
+              assignment: assignment.name || assignment.slug,
+              classroom: `${org}/${classroom}`,
+            }}
+            components={{
+              assignment: (
+                <span dir="ltr" className="font-semibold text-base-content" />
```
rongxin-liu: The load-bearing `dir="ltr"` bidi isolation is hand-repeated inline at ~20 sites across 4 class variants (`font-semibold text-base-content`, `font-semibold`, `font-bold`, `font-medium text-base-content`). MonoLtr already shows the pattern (a shared primitive with an attribute-locking test).

Perhaps worth a sibling `EmphasisLtr`/`IdentifierLtr` primitive (font-weight as a prop/default) to centralize the `dir="ltr"`? Keeps a future site/refactor from dropping it and reintroducing RTL ordering bugs.

rongxin-liu (approved):
Thank you. Nice work!

PR #353, opened 2026-07-21
feat(gh-teacher): add additional details to autograded logs. 

### LINKED ISSUE #260: feat: print failure details to autograde runner logs

Autograder / grading

Currently there is no way to obtain failure details from autograde runner logs. It shows only that "something failed" but not the details:

Image

Also there is no clear indication in the name of jobs as to which job actually runs the test cases/grading.
They are all named generically `Run set -euo pipefail`.

### Proposed solution

Output failure details to the runner log, so that they appear both in submission release and in log.
And also add clear name like "Grade details" to the one job that will contain the details.

### Alternatives considered

_No response_

### Additional context

Previously with GitHub Classroom I often used `diff --color=always ...` to provide a more meaningful failure diffs for students both during autograding on GitHub and locally on their machines.
Runner logs show ANSI coloring very well.

Image

However, currently failure details are shown in markdown for submission and are rendered as plain text, which looks even worse if I try to add color.

Image

### Before submitting

joshkaplan-dev:
## Summary

Failure details for declaratively-graded submissions were only visible in the submission release; the workflow log showed a single summary line, and every step in the Actions UI rendered as "Run set -euo pipefail". Debugging a failed run from the Actions tab — the natural place to look — meant a detour to the release, with no indication of which step even ran the grading.

Three changes, one per commit:

1. **Name every workflow step** in `autograde-runner.yaml`. The grading step is now "Grade details", as requested in the issue. Step ids are unchanged, so all `steps.<id>.outputs` references keep working. The student-repo shim is untouched (it's frozen at accept time; the reusable workflow propagates via its `@main` reference).

2. **Unified diff for `exact` io test failures.** A failing exact-comparison test previously dumped the full expected and actual outputs as separate blocks (up to 2000 chars each). It now renders a `difflib` unified diff that pinpoints the divergent lines. Inputs are stripped to mirror `compare_output`'s whitespace-tolerant semantics, so the diff never flags whitespace the comparison ignores. `included`/`regex` failures keep the verbatim blocks — a line diff against a fragment or pattern is noise. This improves the release body on its own, and diff lines are self-describing (`-`/`+`/`@@`), which lets the log renderer colorize them without extra data.

3. **Print a per-test report to the workflow log** (declarative grader only — custom autograders already stream their stdout). One PASS/FAIL line per test, then one collapsible `::group::` per failing test containing its detail, ANSI-colored (green/red verdicts; red/green/cyan diff lines). Color is gated on `GITHUB_ACTIONS=true` and honors `NO_COLOR`, so local runs and pytest-captured output stay plain. Detail lines are indented two spaces so student-controlled output can never start a line with a workflow command (`::error::`, `::stop-commands::`, …) — GitHub only interprets these at column 0. This is the log-surface analogue of the existing `_fence` guard on the Markdown surface. The release body is also mirrored to the run's Summary page via `GITHUB_STEP_SUMMARY`, on both the declarative and custom-autograder paths.

The log report for a failing submission looks like:

    PASS  program runs  (1/1)
    FAIL  greets by name  (0/3)
    ::group::FAIL: greets by name
      exit 0; comparison=exact
      --- expected
      +++ actual stdout
      @@ -1,3 +1,2 @@
      -Hello, world!
      -goodbye
      +Hello world
       done
    ::endgroup::

Rollout: `runner.py` is fetched from Pages at run time, so commits 2–3 reach every classroom on the next config-repo publish; the workflow step names propagate immediately. No student-repo changes, no schema changes (`detail` was already stripped from `result.json` before this PR).

Verified end-to-end against a test org: named steps and colored collapsible failure groups in the Actions log, score table on the run's Summary page, and the unified diff in the release body.

## Follow-up (second review round)

A second review pass surfaced four items, all addressed in the latest commit:

1. **P1 — non-UTF-8 body crashed a graded run.** `append_removed_files_note` read `release-body.md` with a strict decode; a custom autograder's non-UTF-8 body raised `UnicodeDecodeError` (a `ValueError`, uncaught by the `except OSError`), crashing the runner in `main()`'s `finally` after an otherwise-successful grade. Now reads with `errors="replace"` — the same guard the first hardening round applied to the `finalize_result` mirror.
2. **P2 — Summary page misled on non-success paths.** The `$GITHUB_STEP_SUMMARY` mirror lived only on the two success tails, so an infrastructure error or vacuous pass wrote a `release-body.md` that no surface mirrored — the Summary page was empty, or (with removed files) showed *only* a "Removed N file(s)" note, contradicting the wiki this PR ships. The mirror is now centralized in `main()`'s `finally` (`mirror_body_to_step_summary`), so **every** exit path surfaces the final body.
3. **P2 — duplicated verbatim block** in `execute_test` flattened into one branch.
4. **P3 — control-char parity for `test-name`.** `_validate_test_spec` now rejects control chars (mirroring `tests.go` / `tests-v1.schema.json`), with a defense-in-depth strip in `render_log_report` for the column-0 `::group::FAIL: {name}` sink.

New regression tests pin all four (non-UTF-8 removed-files note, error-path summary mirror, exact-only diff, control-char rejection + `test-name` injection guard). The suite's env-isolation fixture now also clears `NO_COLOR` so the color-gate tests are hermetic.

Closes #260

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

  - Skeleton scripts (Python): `python3 -m pytest cli/gh-teacher/autograders_tests -q` (350 passed) and `python3 -m pytest cli/gh-teacher/skeleton_tests -q` (590 passed)
- [ ] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass. *(N/A — no contract change)*
- [x] If I added or changed a CLI command or flag, I documented it in the wiki (`wiki/Autograders.md`: the three result surfaces and the unified-diff behavior)
- [x] My commits follow Conventional Commits


rongxin-liu (approved):
Works great! Thank you!

PR #363, opened 2026-07-22
feat: add submission release assets

### LINKED ISSUE #332: feat: ability to add custom assets to submission release

Autograder / grading

Sometimes it might be useful to upload artifacts as additional assets either to submission release or to workflow artifacts. Currently submission assets will only contain `result.json` besides sources and there is no way to add to those assets.

### Proposed solution

Add customization option to "Advanced settings" in web UI to specify which paths should be uploaded as additional submission assets regardless of passing or failure of autograding. This can be similar to how "Extra apt packages" or "Allowed files" are specified already.

### Alternatives considered

I have an assignment about markup languages (latex, gnuplot) and previously in GH Classroom I just uploaded the produced artifacts (.pdf and .png) as workflow artifacts using separate job after regular autograding. However, now I would need to either use custom runner workflow or duplicate testing logic in separate workflow separate from autograding entirely.

### Additional context

_No response_

### Before submitting

maximkern:
## Summary

Teachers can now attach extra files from the grading workspace to each submission Release. Classroom 50 stores the configured paths in `release_assets` and preserves them when an assignment is edited or reused.

After grading, the runner validates and stages each configured path — rejecting symlinks and out-of-workspace escapes (opened with `O_NOFOLLOW` and copied from a pinned fd, so a swap between check and copy can't escape the workspace) — and the workflow attaches the valid files to the submission Release **at creation time**, alongside `result.json`. Attaching at `gh release create` (rather than a later upload) keeps assets working under **immutable releases**, which reject post-create asset uploads; a re-run deletes and recreates the release to re-attach atomically. A missing or invalid extra file produces a warning without blocking the grading result, and the run emits an `attached N of M` notice so a partial publish is visible.

Validation is hand-mirrored across the JSON schema, the Go validator, `runner.py`, the workflow's inline validator, and the web TypeScript validator (with matching duplicate-path/basename checks). The Go validator lives in its own `release_assets.go`.

Closes #332

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist


rongxin-liu (dismissed):
Looks good overall! I made a few changes to ensure it works within the immutable release constraint.

PR #378, opened 2026-07-23
fix(web): recover accept secret from team description for bare links

rongxin-liu:
## Problem

Visiting a **bare** accept link for a protected classroom (no `?k=`), e.g. `/<org>/protected-classroom/assignments/test/accept`, failed: the page read the secret only from the `?k=` query param, so with no secret it fetched the **unprotected** Pages path (`.../protected-classroom/assignments.json`) which 404s. For a protected classroom, `assignments.json` is only published under the secret-prefixed path.

## Fix

`AcceptAssignmentPage` now falls back to the capability secret stored in the enrolled student's own team description (recovered via `GET /user/teams`) when `?k=` is absent. The link secret still wins when present.

- Already-enrolled student on a bare link → secret recovered → protected path fetched → accept works.
- Non-member (no team, no secret) → still falls back to the unprotected path and 404s, which is correct (they must use the full `?k=` invite link).

The secret lookup is the shared `useClassroomSecret` hook (also used by `StudentAssignmentList`), and its `GET /user/teams` read is skipped when `?k=` already supplies the secret, so the common path avoids the extra request.

## Tests

Added 3 cases covering secret precedence: `?k=` wins, team-description fallback, and no-matching-team → no secret. `tsc -b`, eslint, prettier, and vitest all green (28 tests across the affected suites).

### DIFF
--- a/web/src/components/org/StudentAssignmentList.test.tsx
+++ b/web/src/components/org/StudentAssignmentList.test.tsx
@@ -51,6 +51,10 @@ vi.mock("@/hooks/useGetMyOrgRepos", () => ({
 }))
 vi.mock("@/hooks/useStudentClassrooms", () => ({
   useStudentClassrooms: () => studentClassrooms(),
+  useClassroomSecret: (_org?: string, classroom?: string) =>
+    studentClassrooms().classrooms.find(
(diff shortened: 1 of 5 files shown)


rongxin-liu:
Closing in favor of a fresh PR from a clean branch (mergeability check was stuck).

PR #510, opened 2026-08-05
fix(web): give the language-pack URL input an accessible name

rongxin-liu:
## What

Adds an `aria-label` (`language.urlLabel`) to the language-pack import URL `<input type="url">` in `LanguageSwitcher`. It previously carried only a `placeholder`.

## Why

A placeholder is not an accessible name (WCAG 2.2 **4.1.2 Name, Role, Value** / **3.3.2 Labels or Instructions**): it disappears on input and many screen readers don't announce it as the field's name. The sibling share-URL input already uses `aria-label`; this brings the import field to parity.

## Audit note

This landed from a static WCAG audit of the web codebase against the guidelines. Findings:

- jsx-a11y ratcheted rules: **0** violations (the earlier ratchet holds).
- `` without `alt`: **0** (all 8 sites already labeled).
- Icon-only buttons without an accessible name: **0**.
- The one raw `<svg>` (progress ring) is correctly wrapped in `<span role="img" aria-label>`.
- Placeholder-only inputs: **1** — fixed here; an app-wide sweep confirms **0** remain.

So this is the only genuine static gap; the machine-detectable layer is otherwise clean. Screen-reader / keyboard verification remains the human manual-assessment pass (unchanged).

## Verification

`npm run check` green — 0 errors, 3222 tests pass, i18n audit accepts the new key.

### DIFF
--- a/web/accessibility/manual-assessment.md
+++ b/web/accessibility/manual-assessment.md
@@ -1,0 +1,144 @@
+# Manual accessibility assessment — WCAG 2.2 AA
+
+Committed assessor checklist for the WCAG 2.2 success criteria that automated
+checks can't establish. These need a human keyboard + screen-reader pass; work
+down each section and record the verdict in the VPAT model.
+
+## How to record a result
+
+For each SC below, after assessing it, edit its entry in
+`web/src/util/vpatModel.ts`:
+
+- set `status` to `supports`, `partially`, or `doesNotSupport`;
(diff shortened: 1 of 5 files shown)


rongxin-liu:
Folded into #513 (the language-pack URL aria-label fix is cherry-picked there). Consolidating the a11y work into one PR on main.

PR #509, opened 2026-08-05
feat(web): scaffold the manual a11y assessment and add an accessibility statement

rongxin-liu:
## What

Sets up the human keyboard + screen-reader assessment (WCAG 2.2 U11) as a fill-in-the-blanks task, and ships the public accessibility statement.

- **Assessor checklist** (`web/accessibility/manual-assessment.md`) — one section per still-`notEvaluated` SC (19 today), each with what "supports" means, keyboard steps, and VoiceOver/NVDA steps, plus the primary flows to exercise (login / accept / submit / roster / org settings).
- **Progress guard** (`vpatManualAssessment.test.ts`) — keeps the checklist in lockstep with the VPAT model: it must document exactly the model's still-outstanding SC set. As a human records a `manual` verdict in `vpatModel.ts`, that SC drops out and its checklist section must be removed, or CI fails. Makes remaining work visible and partial progress landable.
- **Accessibility statement** — a Statement tab on `/accessibility` with the conformance target (WCAG 2.2 AA), how it's evidenced (automated + manual), known limitations, and a feedback path (discussion #493). All prose via `t()`.

## What this deliberately does NOT do

It does **not** produce the assessment verdicts. Those need a human keyboard + screen-reader pass an agent can't perform, and the VPAT design forbids fabricating `manual` verdicts — so the 19 SCs stay `notEvaluated`. This PR is the scaffold that turns the manual pass into "work down the checklist, flip each SC in the model." No new `manual`-verdict wiring was needed: the model/guard/renderer already ingest `manual` evidence.

## Verification

`npm run check` green — 292 test files / 3222 tests, 0 lint errors, prettier clean, i18n audit PASS (statement prose all `t()`-backed). Confirmed no verdicts changed (still 19 `notEvaluated`). Progress guard verified to bite bidirectionally on checklist↔model drift.

Plan: `docs/plans/2026-08-05-004-feat-web-manual-a11y-assessment-scaffold-plan.md`. The manual assessment itself is the standing human follow-up the checklist enables.

PR #511, opened 2026-08-05
feat(web): interactive dev-only WCAG assessment mode

rongxin-liu:
## What

Turns the manual WCAG checklist into an interactive dev-only tool. Run `npm run dev`, open `/assess`, click through the outstanding success criteria, pick a verdict + type a remark, and it saves straight back to the repo.

## Stacked on #509

This builds directly on the manual-assessment scaffold in #509 (the checklist + drift guard), so it targets that branch. **Merge #509 first**, then this retargets cleanly to `main`.

## How it works

- **JSON overlay** — human verdicts live in `web/src/util/vpatVerdicts.json`, merged into the VPAT model via `applyVerdicts()`. It only lands on `notEvaluated` rows and *throws* if a verdict would overwrite an automated/contrast row or an unknown id, so the tool can never silently overclaim. `vpatModel.ts` stays the readable spine.
- **Data-rendered checklist** — `manualAssessmentDoc.ts` holds per-SC assessor guidance and a pure renderer that rebuilds `manual-assessment.md` from the model's outstanding set. A freshness test pins the committed doc to the renderer byte-for-byte; a recorded verdict drops its section automatically (the drift guard stays green).
- **Dev-only write bridge** — `assessmentApiPlugin` (`apply: 'serve'`) serves `GET /_assess/data` and `POST /_assess/save`, writing the JSON overlay + regenerating the checklist. The app is client-side only, so this is the local dev bridge; it never exists in a production build.
- **Dev-only route** — `/assess` `beforeLoad` redirects to `/` unless `import.meta.env.DEV`. Verified: in the production bundle the guard dead-code-eliminates to an unconditional redirect and no `/_assess` asset is emitted.
- **AI polish** — the checklist header instructs assessors to ask the agent to standardize remark wording after a session.

## Safety

- No production surface: serve-only plugin + DEV-gated route (both verified against a real build).
- No path traversal: writes are two fixed `path.resolve` constants; the user-supplied `id` never reaches a filesystem path.
- Loopback-only: `/_assess/*` returns 403 for any non-localhost `Host`, so `vite --host` can't expose the unauthenticated write endpoint.
- Consistent with AGENTS.md: prod stays client-only (this is local dev-tooling), util remains a leaf, and the i18n audit exempts the dev-only page.

## Review

`ce-code-review`: **Ready to merge** (no P0/P1/P2). Two P3 nits and the LAN-exposure residual were all fixed in the final commit (loopback Host guard, `GET` try/catch, `clear`-path id validation).

## Verification

`npm run check` green (0 errors, 3229 tests). End-to-end exercised live: save/clear round-trip, validation guards (bad status/unknown id/automated-row/non-string id all rejected), loopback enforcement, and pristine repo after clear.


rongxin-liu:
Superseded by #513 (consolidated onto main; the stack was unmergeable).

PR #531, opened 2026-08-07
feat: per-assignment submission triggers — modes and milestone tags (#477)

### LINKED ISSUE #477: feat: configurable submission trigger (custom tag vs. every push)
## Summary

Today the autograder shim in every student repo triggers on **every push to the default branch** (plus `submit/*` tags), so every commit grades and burns GitHub Actions minutes. At large-class scale (e.g. 500 students) this exhausts the org's Actions quota and inflates the submission count.

This feature lets a teacher choose, **per assignment**, a **submission mode**:

- **`every-push`** — current behavior (the default; unchanged).
- **`tag`** — grade only when a student pushes a **teacher-configured tag** (e.g. `submission`, `v*`, `release-*`), instead of on every commit.

**Why it matters.** It's the primary cost lever for large cohorts, and it makes "what counts as a submission" an explicit teacher decision rather than an accident of how often students push.

## How it works (direction, not implementation)

- **`submission_mode` (+ tag patterns) on the assignment.** The mode and the configured tag pattern(s) live in `assignments.json` (`classroom50/assignments/v1`), absent = `every-push`, so every existing assignment is unaffected. Editable after creation, and threaded across all cross-tool mirrors (schema, Go, Python, TS) additively so older readers tolerate it.
- **The trigger lives in the shim, generated from the mode.** A workflow's `on:` block is evaluated by GitHub *before* any job runs, so it can only come from the shim file inside each student repo — not from the manifest at grade time. `every-push` keeps today's branch + `submit/*` trigger; `tag` mode drops the branch trigger so ordinary commits don't grade, emitting only the teacher's tag pattern(s).
- **The tag config is the source of truth for "what is a submission."** Today a submission is defined solely by the `submit/*` tag prefix, hardcoded across the runner, the score collector, regrade, and the web submission reads. Arbitrary tags require that whole pipeline to consult the assignment's configured pattern(s) instead — via a single shared "is this tag a submission?" matcher.
- **How a student submits in `tag` mode: plain `git`, no tooling required.** A student commits, tags the commit with a matching name, and pushes the tag — `git tag submission && git push origin submission`. The tag push is what triggers grading. This is the **baseline contract**: `gh student submit` and the web "submit" are conveniences that automate the same tag-and-push, but the feature must not assume them — plain `git` alone must work end-to-end.
- **`submit/*` is legacy, not required — and its use is discouraged.** Custom tags are the **preferred** approach going forward. `submit/*` is kept working only as a **backward-compatibility default** (when an assignment has no tag config) so existing classrooms keep grading unchanged. That legacy behavior is **planned for eventual removal** once every assignment carries explicit tag config (see Migration below).

> **Important:** this is **not** a trigger-narrowing tweak that can ship on one side alone. Because the runner, the score collector, and regrade are all built around the `submit/*` prefix, `tag` mode only works as **one coordinated change** across the schema, the reusable autograde runner, the score collector, and regrade — plus the submit tooling, which is updated to automate the plain-`git` tag push (a convenience, not the contract). Treat it as a single cross-tool effort.

### 1. Per-assignment submission mode + configurable tag(s)

Let a teacher pick `every-push` or `tag` per assignment, and in `tag` mode configure the tag pattern(s) that count as a submission (defaulting to a sensible built-in, `submission`). Support the GitHub Actions [tag filtering](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushbranchestagsbranches-ignoretags-ignore) surface (globs like `v*`, `release-*`), validated before write. Keep the door open — additively — for future extensions such as `tags-ignore` exclude patterns and custom branch triggers. `every-push` stays byte-identical to today.

### 2. Make `tag` mode actually grade — and count — a submission

Because the current submission pipeline rides the `submit/*` prefix end-to-end, `tag` mode requires coordinated changes so a tagged submission both grades and lands in the gradebook:

- The shim's `on:` block generated from the mode.
- **Plain `git` is the submission baseline** — a student can commit, tag with a matching name, and push the tag with no Classroom 50 tooling. `gh student submit` and web "submit" **automate** the same tag-and-push as a convenience, never a requirement.
- The **runner** accepts that tag as a submission (generalized from the `submit/*`-only gate) and records the actual tag.
- The **score collector**, **regrade**, and the **web live-submission reads** all recognize a submission by the assignment's configured pattern, not the hardcoded prefix.

A single shared matcher should back all of these so they never disagree — a divergence would silently drop submissions from the gradebook. `submit/*` is kept working only as a backward-compatibility default (see the deprecation stance above and Migration below); custom tags are preferred. Document the student-facing consequence: after a teacher changes the trigger, a student's existing clone is stale and must be re-pulled before the next submit.

### 3. Update the shim in existing student repos (single + bulk)

A mode change only takes effect once the shim is rewritten in already-accepted repos. Provide a way to update the shim in one repo and in **bulk** across a cohort, reusing the existing enumerate-repos-and-apply machinery, idempotent (skip repos already current) with per-repo progress and failures. The rewrite must **not** itself trigger a wave of grade runs — the runner skips the bot shim-update commit. **Custom (non-default) autograders are excluded** from the mode feature, so a teacher-authored shim is never rewritten.

## Migration to the new format (retire the legacy default)

The backward-compatibility path (treating an assignment with no tag config as `submit/*`) is a temporary bridge, not the end state. To retire it cleanly:

1. **Migrate `assignments.json` additively.** Provide a migration that rewrites each existing assignment to the new format by writing **explicit tag configuration** — for an assignment with no config today, that means recording the current `submit/*` behavior explicitly (e.g. `submission_mode` reflecting the legacy trigger with the `submit/*` pattern made explicit). This is a pure config rewrite (no student-repo changes) and is idempotent: an already-migrated assignment is left alone.
2. **Preserve behavior exactly during migration.** Because migration writes the legacy `submit/*` behavior *explicitly*, a migrated assignment grades and collects identically to before — the only change is that "what counts as a submission" is now stated in the manifest rather than implied by a hardcoded default.
3. **Remove the backward-compatibility logic afterward.** Once every assignment (across every classroom) carries explicit tag config, the "absent config ⇒ `submit/*`" fallback can be deleted from the shared matcher and all its call sites, leaving a single, explicit code path. That removal is a follow-up, gated on migration coverage.

This lets us ship arbitrary tags now without breaking existing classrooms, converge on one explicit format, and drop the legacy branch later with confidence.

## Scope

- [ ] `submission_mode` (+ tag patterns) schema field and all cross-tool mirrors, with parity tests
- [ ] A single shared "is this tag a submission?" matcher (one per language), parity-pinned, with a temporary `submit/*` backward-compatibility default (slated for removal post-migration)
- [ ] A migration that rewrites `assignments.json` to the new format with explicit tag config (legacy `submit/*` behavior preserved explicitly), idempotent; backward-compat logic removed as a follow-up once coverage is complete
- [ ] Shim `on:` generation driven by the mode (`every-push` unchanged)
- [ ] `tag` mode grading end-to-end via **plain `git`** (tag + push), with `gh student submit` / web submit automating the same as a convenience; runner accepts + records the tag
- [ ] Collection contract: score collector, regrade, and web live reads recognize submissions by the assignment's patterns (never silently dropping previously-collected submissions when the config changes)
- [ ] Runner skips the bot shim-update commit (so bulk updates don't self-grade)
- [ ] Single + bulk shim-update across existing repos (default-autograder assignments only)
- [ ] Assignment create + settings UI (web + CLI), with a wildcard-collision caution and a "changing the mode rewrites shims across student repos" warning
- [ ] Student-facing guidance leading with the plain-`git` tag recipe for `tag` mode, plus "the shim changed — re-pull before your next submit"

### Deferred / extensibility (not now)

- `tags-ignore` exclude patterns (additive later; teachers want an allow-list first)
- Custom branch triggers (a future mode value / additive field)
- Ambient "N repos still on the old trigger" drift-detection UI

### Notes / open questions

- **Collision is an accepted consequence of arbitrary tags:** a broad pattern like `v*` grades on any matching tag a student pushes, including a real release tag they cut for unrelated reasons. The teacher-facing UI should caution on wildcard patterns and steer toward the specific `submission` default.
- **The re-tag-an-earlier-commit workflow** (issue's headline: a student tags an older commit; the teacher sees that commit's timestamp) should be confirmed end-to-end.

### Related work & building blocks

Supporting context to help shape this feature — existing building blocks to reuse and adjacent levers to keep distinct — grouped by theme.

**Existing "narrow when grading runs" levers (design as a coherent set).** This feature narrows *when* the shim fires; these narrow *whether* the grade job runs. The gates should read cleanly side by side and share the same manifest-field track.

- [#458](https://github.com/foundation50/classroom50/pull/458) (merged) — auto-skips the grade job when *no* autograder is configured. The runner already knows how to no-op cheaply; this is the precedent for a config-driven gate. `submission_mode` should follow the same schema-mirror-in-lockstep track as the existing collapsed-field options (`feedback_pr`/`locked`/`pass_threshold`).
- [#453](https://github.com/foundation50/classroom50/issues/453) (WTCC 6) — a related per-assignment autograding request; same cost-driven motivation as this issue.

**Bulk shim / per-repo rewrite machinery (reuse for the "hard part").** Changing modes means re-committing the shim into every student repo — reuse an existing enumerate-repos-and-apply flow rather than inventing one.

- [#409](https://github.com/foundation50/classroom50/pull/409) / [#434](https://github.com/foundation50/classroom50/pull/434) / [#435](https://github.com/foundation50/classroom50/pull/435) — decoupled Feedback-PR creation and added the bulk open/repair machinery (already cited above); the reference implementation for a batch shim update.
- [#466](https://github.com/foundation50/classroom50/pull/466) / [#469](https://github.com/foundation50/classroom50/pull/469) / [#471](https://github.com/foundation50/classroom50/pull/471) (merged) — the newest per-repo + bulk action flows (assignment-repo access, per-submission actions). Fresh examples of the same enumerate-and-apply pattern and its web/CLI UX.

**Cost & "what counts as a submission" origin (motivation).** The reports and shipped mitigations that seeded this trigger-mode work.

- [#347](https://github.com/foundation50/classroom50/issues/347) — the parent report (missing submissions + submission cost at 500-student scale) this issue is split from.
- Discussion [#461](https://github.com/foundation50/classroom50/discussions/461) (also noted in the comment below) — surfaces the adjacent gap that a submission must produce a `result.json` to be counted. This feature keeps that intact: in `tag` mode the runner **does** run on the submission tag, so a tagged submission still grades and is collected.
- Discussion [#23](https://github.com/foundation50/classroom50/discussions/23) — earliest ask to disable the autograder Action for cost reasons.

**Adjacent submission-timing requests (interacts with the tag/re-tag workflow).** The "which commit is graded, and when" question this feature touches.

- Discussion [#29](https://github.com/foundation50/classroom50/discussions/29) — deadlines with a time component (not just a date).
- Discussion [#199](https://github.com/foundation50/classroom50/discussions/199) — a cutoff date after which access is read-only. Both bear on which commit a `tag` submission points at relative to a deadline.

joshkaplan-dev:
## Summary

Adds per-assignment control over *when* the autograder runs, completing #477 in two parts delivered together: a submission **mode** (every push vs. submit-only) and teacher-named **milestone tags** that trigger grading.

By default every push grades, spending Actions minutes on every work-in-progress commit. This PR adds two levers:

- **`submission_mode`** — `every-push` (default, unchanged) or `tag` (only `submit/*` tag pushes grade; a plain `git push` costs no Actions minutes).
- **`submission_tags`** — optional tag patterns (e.g. `phase1`, `phase2`) that also trigger grading. A student pushing a matching tag gets that commit graded with plain git — no CLI required.

**Milestone tags trigger; `submit/*` records.** A milestone tag can move between commits, so a milestone-tag push *triggers* grading and the runner mints the canonical immutable `submit/<UTC-timestamp>-<short-sha>` tag at that commit and publishes the release there. The collector, regrade, web views, and download are unchanged — all still key on `submit/*`. (This is why the issue's migration step is unnecessary: `submit/*` stays infrastructure, not config.)

**Also in this PR** — the retrofit tooling (`gh teacher assignment submission-mode` + web bulk/per-row actions) that rewrites existing repos' shims; the runner-side stale-shim and foreign-tag suppression; and a few fixes found during live testing:

1. set-latest was dead code (self-compared, never acted) → explicit best-effort claim; latest = most recent submission.
2. Retrofit misreported after a cached pre-write read → reports from the actual commit outcome.
3. Regrade failed on immutable-release orgs (refused delete) → tolerates the refusal, keeps the old release, warns.
4. Milestone-tags field validated only on save → live validation + a comma-vs-newline hint.

**Design deviations from the issue:** no config migration (obsoleted by the triggers-vs-records design) and no default `submission` tag (absent = no milestone tags).

**Deferred (per the issue):** `tags-ignore` / exclude patterns, custom branch triggers, drift-detection UI.

Closes #477

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist


rongxin-liu (approved):

rongxin-liu:
Nice work @joshkaplan-dev! Thank you!

PR #609, opened 2026-08-14
docs(wiki): add August training slides and video, drop signup link

rongxin-liu:
Updates the wiki Home page now that both training sessions have taken place.

- Reword the training note from future to past tense: the sessions are described as having been hosted, with recordings and slides available below (rather than announcing upcoming events).
- Add the August 14 session's slides and recording, mirroring the July entry's format.
- Move both slide decks into a new `wiki/slides/` folder and rename them with `yyyymmdd` dates (`_20260724`, `_20260814`) so each deck is self-describing.
- Remove the registration signup link, since both sessions are now in the past.


brianyu28 (approved):
Thanks!

Issue #456, opened 2026-07-30
WTCC 7: Feature request: Let students select their roster entry when accepting (as GitHub Classroom did) — the teacher can't map GitHub accounts to students otherwise


Talonness:
### Which part would this affect?

Other / not sure

## Which part would this affect?

Roster / student onboarding

---

## The problem

There is no way for a teacher to learn which GitHub account belongs to which student, short of asking every student individually.

When a student accepts an assignment, Classroom 50 writes their row as:

```
username,first_name,last_name,email,section,github_id,role
mquintero,,,,,10482913,student
```

Username, ID, role — and nothing else, because the sync only knows what GitHub knows. Our student information system, meanwhile, gives us names, email addresses and section labels, but no GitHub usernames — students haven't given us those yet.

The two records share no field. Matching is on username at every stage, so a CSV carrying names and emails can only attach details to a row whose username we already know. We tested this three ways: against pending invitations, against an enrolled row, and against an enrolled row that already stored the exact email address in our file. All three reported *"No valid GitHub usernames were found."*

So the only route to a roster that says who anyone is runs through collecting a GitHub username from every student by hand — an assignment submission, a form, an email — and then merging it against our SIS export ourselves.

At our 26-student sections that's an afternoon's annoyance. At 150 students it's a different proposition, and it produces nothing the institution didn't already know.

---

## How GitHub Classroom solved this

GitHub Classroom let the teacher upload a roster of arbitrary identifiers — we used student names — with no GitHub usernames required. When a student clicked the assignment link, they were shown the roster and selected their own entry. Classroom linked their GitHub account to that identifier at that moment.

It also let the teacher undo that link. When a student picked the wrong entry, we could detach them from it — the identifier returned to the unassigned pool, and the student could choose again or we could assign it ourselves. We don't recall the exact mechanics and are describing the effect rather than the implementation, but in several terms of use that was the only failure mode we saw, and it never cost more than a couple of minutes.

The teacher never had to know anyone's username. The student, who obviously knows which entry is theirs, supplied the only piece of information the teacher couldn't.

That seems like the right shape of solution: the mapping problem is unsolvable for the teacher and trivial for the student, so it belongs on the student's side.

## Proposed behaviour

1. A teacher uploads a roster CSV with names and sections and **no usernames** — exactly what a student information system exports.
2. Students are invited (by email, as now).
3. On first accepting an assignment, a student is shown the unclaimed roster entries and picks their own.
[...]

Issue #414, opened 2026-07-26
WTCC 3: Roster invitations by email do not retain the email address


Talonness:
### Which part would this affect?

Teacher CLI (gh-teacher)

### Problem or current behavior

**Type:** Feature request  **Priority:** P3 — Low  **Part affected:** classroom50.org

The roster's Add member dialog accepts either a GitHub username or an email address. Adding by email sends an organization invitation but records no name and no section — the dialog states these must be added later by username or by uploading a roster.

Instructors generally do not know students' GitHub usernames before a term begins, so email is the only practical option at the start of a course.

After the student accepts, the roster entry and the submissions table identify them by GitHub username only, with Email and Section columns blank. The invited email address is not retained anywhere visible. GitHub usernames are frequently unrelated to students' real names, so there is no reliable way to match a submission to an institutional gradebook other than manual lookup.

### Proposed solution

Retain the invited email address on the roster entry after acceptance, and display it in the submissions table alongside the GitHub username.

### Alternatives considered

**Upload a roster CSV with names and usernames.** Requires knowing students' GitHub usernames in advance, which instructors generally do not at the start of a term.

**Rely on institutional email addresses being name-shaped.** Works only where addresses follow a predictable pattern, and still requires the address to be visible somewhere after acceptance — which it is not.

**Backfill names by username after acceptance.** Possible, but this is the manual reconciliation the request is trying to eliminate: 26 students per organization across three organizations.

### Additional context

Reproduction:

1. In a classroom roster, use Add member and supply an email address rather than a GitHub username.
2. Have the invited account accept the organization invitation and accept an assignment.
3. Open the assignment's submissions table and the roster; observe the student is identified by GitHub username only, with Email and Section blank.

### Before submitting

- [x] I searched existing issues and this is not a duplicate.

rongxin-liu:
Thanks for opening the issue.

This is actually an intentional design decision. We do not retain email addresses in the roster CSV file because we cannot reliably associate a GitHub username with the email address used for the invitation. GitHub simply does not provide this mapping. If we included email addresses, the roster CSV file could end up containing rows with email addresses but no associated GitHub usernames.

This would be achievable if classroom50.org had a server running, but it does not.

rongxin-liu:
[...]

PR #631, opened 2026-08-16
feat(web): retain invited emails and recover them on acceptance

rongxin-liu:
## The problem

A GitHub org invitation sent to an email address does not retain that address. Once the student accepts, the API reports their login but no longer the email they were invited with, so the teacher has no way to map the new GitHub account back to the person on their class list. There is exactly one moment where that mapping exists and nothing persists it.

## The approach

Each email invite creates a per-invite **secret** GitHub team named `invite-<sha256(classroom \0 email)[:16]>`, whose description holds a PII-minimal `classroom50/invite/v1` record (`{schema, email, classroom}` — nothing else; names and sections belong in `roster.csv`). The invitation carries this team alongside the classroom team, so accepting lands the student on both. A later reconcile finds the team's one regular-role member, pairs their account with the recorded email, writes it to `roster.csv`, and deletes the team. The team is `secret` and the write is fail-closed, so an invited email is never exposed to other org members.

The invited address is also written to `roster.csv` immediately as an email-only pending row, so it is retained from the moment the invite is sent rather than only on acceptance. That row lives exactly as long as its invite does: the reconcile fills in the account identity when the student accepts, and removes the row if the invite is cancelled, expires, or is garbage-collected.

## One reconcile, one commit

`reconcileRoster` is now the single roster writer, shared by the roster page's Sync action and the owner-visit classroom reconcile: **collect** classifies the invite teams (read-only), then **one commit** folds recovered identities, drops dead email rows, appends missing members and reconciles roles/ids, then **finalize** deletes the recovered teams only after that commit lands. This replaces the previous per-invitee backfill commits plus a separate sync commit.

Destructive steps are gated so a degraded read can never be mistaken for absence: row removal requires a trustworthy invite-team pass **and** confirmation against GitHub's current pending invitations, read inside the retried commit closure (the collect snapshot predates the roster read, so an invite sent in between must not be reaped). An unreadable invitation list keeps every row. Teams are reaped only behind a 24h age guard when no pending invitation maps to them, and cancel/dismiss tears a team down immediately.

## Limitations

The retained address is the one the teacher **invited**, not necessarily the email registered on the student's GitHub account. Surfaced in the invite form's helper text and the teacher guide.

## Follow-up required before or with this merge

The Go teacher CLI's strict roster reader rejects a row with an empty `username` and aborts the whole parse, so `gh teacher roster list` errors while an email invite is pending. Aligning the CLI reader (mirroring the web keep-rule `username || github_id || email`, plus a parity test) is a separate PR and should land first or together.

## Test plan

- [x] `npm run check` green — tsc, eslint, prettier, dependency-cruiser, 3865 unit tests
- [x] Verified live on a real org (`classroom50-summer-dev`): invite writes `,,,student@example.edu,,,student`; on acceptance the same row upgrades in place to `rliu50,,,student@example.edu,,127826836,student` with no duplicate; the metadata team is deleted; two commits total for the whole lifecycle
- [x] Verified the mechanism directly: GitHub adds the invitee to a freshly created secret team via `team_ids`, and keeps org owners as team maintainers (hence the `role=member` read)
- [ ] Browser accessibility suites (`*.browser.test.tsx`) run in CI only — need `npx playwright install chromium` locally; no layout changes in this branch

Towards #414. Also removes the need for #456 — teachers can reconcile GitHub accounts against their SIS by email, so students never need to self-select a roster entry.

### DIFF
--- a/schemas/invite-v1.schema.json
+++ b/schemas/invite-v1.schema.json
@@ -2,7 +2,7 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "https://raw.githubusercontent.com/foundation50/classroom50/main/schemas/invite-v1.schema.json",
   "title": "classroom50/invite/v1",
-  "description": "Compact record stored in the DESCRIPTION of a per-invite `secret` team (`invite-<hash>`, where <hash> is a 16-hex-char SHA-256 prefix of `<classroom>\\u0000<lowercased-email>`) created when a teacher invites a student to a classroom by email. The team is attached to the org invitation's team_ids alongside the classroom team, so on acceptance the invitee becomes a member of BOTH. On the teacher's next visit a reconcile pass finds the team's single non-owner member (the accepted invitee), pairs their login/id with the `email` recorded here, backfills roster.csv, and deletes the team. This exists only to bridge the ONE moment nothing else can: after acceptance GitHub reports the login but no longer the invited email. Readable only by team members and org owners (the team is secret), so an invitee's email is never exposed to other students. The record is temporary storage, never a source of truth — enrollment and role stay team-driven, and teacher-entered roster.csv values win on conflict. Because the accepted invitee can edit their own team's description, a reader treats it as UNTRUSTED input: schema-validate it, require `email` to match the team-name hash, and treat name/section as best-effort display values. Keep the encoded JSON well under ~250 chars (GitHub's team-description limit): a writer over budget drops the optional display fields first, always preserving `schema`, `email`, and `classroom`. `additionalProperties` is true only to TOLERATE a field a newer binary adds; unknown fields are not preserved on a rewrite. Web-only today (email invites exist only in the web app); no Go/Python mirror yet.",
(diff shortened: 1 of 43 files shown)


mheckner:
Hello,

First of all, thank you for reacting so quickly to my questions and comments yesterday.

As I understand it, the new roster-sync feature works like this: adding a student email to roster.csv creates a pending row and a secret invite-<hash> team holding just that address; once the student accepts the invite, the sync picks up their username from that team, writes it into the pending row, and deletes the invite team. That would solve exactly the problem I had — really great feature!

However, when I tested it, it didn't quite work. Here's what I did and saw:

I uploaded a roster file containing a single email address that was not currently a member of the org (I had removed them beforehand).
GitHub sent the invitation email — but it only mentioned the classroom team, not an invite team.
I checked the org and saw no invite-<hash> team was created at all.
roster.csv was never updated with a pending row for that address.
Once the invitation was accepted, the student did get added — but the next roster sync only filled in their GitHub username, not the pending-row/team linkage you described.

So it seems like the invite-team creation step either isn't running or is failing silently for me, and the roster only picks up the username after acceptance, not the intermediate pending state.

A few questions:

Is there something I need to configure or update on my end that I might be missing? The web UI doesn't show any workflow /action updates that I might to carry out.
Would adding you to my org help you debug this?
As a fallback for identifying students (independent of this bug): would it be possible to also pull the public "name" field from the GitHub profile? I know students can put anything there, but we could enforce it by saying grading depends on being identifiable.

Thanks again for the great work — I really appreciate it.

Best regards,
Markus

rongxin-liu:
This feature is not 100% complete and has not yet been released to production, but you can try it at https://preview.classroom50.org to see how it behaves.

We are still working on completing the feature so that the other roster-processing logic can properly handle email invitations.

mheckner:
It works perfectly in preview, thanks! Meets all my requirements! :)

rongxin-liu:
@mheckner Nice, glad to hear that!

fulcorno:
I would like to thank you for the quick response and resolution on this issue. Having a course with 230 students, it really saves my day (I invite the official university email, from which I can recover any other data).

A small suggestion: It would be great it this functionality worked also by uploading a CSV roster (with email, but no username), so that we can pre-fill also other info and trigger the invitation process. But I understand that this info should be "parked" in the secret invite, and it would probably be against your design decisions.

rongxin-liu:
@fulcorno As part of the update, the app would also support email-only CSV uploads and bulk sending of invites accordingly. However, if the invites are canceled or expire, these email entries (without an associated GitHub username or ID) will be automatically removed.

PR #635, opened 2026-08-17
fix(web): keep an invite record an org owner accepted

rongxin-liu:
The last silent data loss in the email-invite feature (the remaining P2s are hygiene). Follow-up to #631, #632, #634.

## What was lost

Recovery reads an invite team's members with `?role=member`, because GitHub keeps the creating teacher — and every org owner — on the team as a *maintainer*, and the accepted invitee is a plain member. That works until the invitee **is** an owner: GitHub auto-promotes them, so they never appear in that read and the team looks like nobody ever accepted.

The consequences chained: the 24h GC reaped the team as abandoned, and the next sync then dropped its roster row as dead — so the invited address disappeared with no trace, for someone who really had joined. The bulk path already avoids creating a metadata team for teacher-role invites, but that doesn't help a student promoted to teacher between invite and reconcile, which is the realistic path into this.

## What changed

Before reaping, the pass reads the team's maintainers. The creating teacher is one, so a **second** maintainer means somebody accepted. The mapping still can't be recovered — an owner-invitee is indistinguishable from the inviting teacher by team membership alone — so rather than guess an identity, the record is kept: the team survives, its email counts as live so the sync leaves the roster row alone, and the address is returned in a new `ownerAccepted` list for a caller to surface. The teacher keeps seeing the address and can finish the row by hand or cancel the invite.

The extra read is on the GC decision path only, so an ordinary reconcile pass doesn't pay for it (pinned by a test).

## Why not recover it properly

Two ways were considered and rejected. Reading maintainers and taking the non-teacher one breaks as soon as a classroom has more than one owner, and would happily map an address onto the wrong account — worse than keeping the row. Skipping the metadata team for anyone who *might* become an owner isn't knowable at invite time. Keeping the record is the honest outcome: no data is destroyed, and the one thing the automation can't determine is left to the teacher.

## Still open (hygiene, not loss)

Archiving a classroom leaves invite teams behind — deliberately not auto-purged, since archive is reversible and unarchive would want them back; it wants a notice plus the existing cleanup button. Nothing cancels the org's pending email invitations on delete/archive/teardown (GitHub expires them). The CLI can read a pending row but not remove/update it. And `matchesRosterRow`'s comment still claims "there are no username-less rows", which this feature falsified.

## Test plan

- [x] `npm run check` green — tsc, eslint, prettier, dependency-cruiser, 3889 tests
- [x] A second maintainer keeps the team, reports the email in `ownerAccepted`, and keeps it live so the row survives
- [x] Only the creating teacher as maintainer still reaps, as before
- [x] A young pending team pays for neither the invitation read nor the maintainer read
- [ ] Browser accessibility suites run in CI only (need `npx playwright install chromium` locally); untouched here

### DIFF
--- a/web/src/domain/students.test.ts
+++ b/web/src/domain/students.test.ts
@@ -2867,6 +2867,7 @@ describe("syncRosterFromTeam — identity-only backfill", () => {
     liveInviteEmails: new Set<string>(),
     trusted: true,
     deletedStale: 0,
+    ownerAccepted: [],
     ...over,
   })
 
--- a/web/src/domain/students/inviteRecoveries.test.ts
+++ b/web/src/domain/students/inviteRecoveries.test.ts
@@ -4,6 +4,7 @@ import { describe, expect, it, vi, beforeEach } from "vitest"
 const listInviteTeams = vi.fn()
 const readInviteTeam = vi.fn()
 const deleteInviteTeam = vi.fn()
+const listInviteTeamMaintainers = vi.fn()
 const listTeamMembers = vi.fn()
 const listOrgInvitations = vi.fn()
 const resolveClassroomTeamSlugs = vi.fn()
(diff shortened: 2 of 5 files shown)


rongxin-liu:
Superseded by a better fix. Rather than *detecting* an org-owner invitee with a maintainer-count heuristic and keeping an unrecoverable record, the invite team now drops the acting teacher at creation, so it holds no teacher and any member of any role is unambiguously the invitee — the owner case recovers fully instead of merely avoiding data loss.

Verified live that an org owner can delete their own team membership (204, team then reads as genuinely 0-member) and that a non-member owner can still read, patch, delete, and enumerate the secret team.

That also lets the `role=member` filter, `listInviteTeamMaintainers`, and the bulk path's admin-role skip all go away — so teacher-role email invites gain email retention, which they don't have today.

Issue #654, opened 2026-08-18
migration banner: an absent submission_mode is the schema's every-push default, not a legacy marker — migrating writes "tag" and silently stops push grading


graphics80:
### Which part is affected?

Web app (classroom50.org)

### What happened?

The classroom migration banner (`MIGRATION(v1.28)`) treats an **absent** `submission_mode` as "legacy, needs migrating", and its migration writes `"submission_mode": "tag"` onto every such entry. But per the schema, an absent field is not legacy — it is the mandated wire default, and it means `every-push`:

> On the wire it is collapsed like `feedback_pr`: writers omit `"every-push"` (the CLI normalizes it away), so readers must treat an absent field as `"every-push"`; accept an explicit `"every-push"` too.
> — `schemas/assignments-v1.schema.json`, `submission_mode`

So the migration does not normalize the file, it **flips the mode**: `every-push` → `tag`. Since `submission_mode` is baked into each student repo's shim at accept time, every repo accepted after the migration gets a shim with only `tags: ["submit/*"]`. A student's plain `git push` then produces no run, no release and no grade — silently, with no error anywhere.

Three consequences of one click:

1. **New accepts stop grading on push.** No warning in the modal; its text says *"It does not change any score, due date, or student's access, and you can keep working as before."*
2. **The file and the existing repos disagree.** The migration writes only `assignments.json` and does not retrofit shims (unlike `gh teacher assignment submission-mode`), so already-accepted repos keep grading on push while the manifest says `tag`. Two shim generations run in parallel, and which one a student got depends on the minute they accepted.
3. **The submissions page starts counting the wrong thing** for exactly those repos: `useDetectedSubmissions` is passed `mode: assignmentInfo.submission_mode`, so it counts tags for repos that in fact grade on every push.

The code comment in `normalizeEntryForMigration` (`web/src/domain/assignments/createEdit.ts`) justifies `"tag"` with counting parity — a pre-1.28 submission was a `submit/*` release — but only reasons about the submissions page. It does not account for `submission_mode` also being the accept-time shim trigger (`accept.ts` → `defaultAutograderWorkflow`).

**The banner also cannot be gotten rid of**, because `gh teacher assignment add` deliberately writes the same absent field:

```go
// Normalize the wire default away so an every-push assignment's
// entry stays byte-identical to one written before the field
// existed.
submissionModeVal := strings.TrimSpace(submissionMd)
if submissionModeVal == contract.SubmissionModeEveryPush {
[...]

PR #681, opened 2026-08-21
fix(web): stop the classroom migration from stamping grading: auto (#680)

### LINKED ISSUE #680: web: classroom migration stamps an explicit grading: { mode: "auto" } that every other writer omits
### Which part is affected?

Web app (classroom50.org)

### What happened?

The classroom migration (`migrateClassroomAssignments` → `normalizeEntryForMigration` in `web/src/domain/assignments/createEdit.ts`) writes an explicit `grading: { mode: "auto" }` onto every legacy entry that has no `grading` block.

No other writer does that. The CLI has no `--grading` flag and omits the pointer (`json:"grading,omitempty"`); the web app's own `buildAssignmentEntry` omits the block for plain auto ("mirroring the CLI's omitempty"). The schema reads an absent `grading` as `auto` ("ABSENT reads as `auto` so existing files are unchanged") and documents "writers omit the default" for every other defaulted field; only `submission_mode` is exempted, deliberately, because its presence is the migration signal.

**What this causes**

Nothing breaks at runtime — every reader resolves `grading?.mode ?? "auto"` — but the stray block has real costs:

1. **Fabricated intent.** The schema defines `grading` as "the teacher's grading intent as a first-class choice the GUI surfaces". Absent means *no choice recorded*; `{ mode: "auto" }` means *the teacher chose auto*. The migration invents the second from the first for every legacy assignment, so after migration nobody (GUI, CLI, a future "explicitly configured?" check) can tell a recorded choice from a stamped default.
2. **Phantom diffs in the config repo.** The migration commit adds `grading` to every legacy entry. Merely opening any of those assignments in the web app and clicking Save — changing nothing — removes it again: `editAssignment` has no no-op short-circuit, it always rebuilds the entry via `buildAssignmentEntry`, which owns `grading` and omits auto. So the teacher gets one "Edit assignment" commit per assignment whose only content is the block disappearing. Not a ping-pong (once gone it stays gone), but across a classroom with N assignments that is N commits the teacher never asked for, and `git log -p assignments.json` shows `grading` appearing and vanishing with no decision behind either.
3. **Mixed fleet.** Classrooms migrated on web-v1.28 … current carry the block; entries edited via the CLI afterwards keep it (the CLI carries an existing pointer forward), entries edited via the web lose it, never-migrated or post-fix files never had it. Same meaning, three different on-disk shapes across one org, which makes diffing classrooms or answering "why does hw1 have `grading` and hw2 not?" needlessly confusing.
4. **Convention broken by one writer.** The schema says "writers omit the default" for every defaulted field and exempts only `submission_mode` (its presence is the migration signal). One writer ignoring that means the convention can no longer be relied on when reading a file — e.g. a tool that treats the presence of a defaulted field as "explicitly set" is now wrong for migrated files.
5. **Misleading code comment.** The function comment calls the result the "canonical shape"; per the app's own create path the canonical shape for auto is *absent*. A maintainer following the comment would re-add the block or assert it in tests.

`isSubmissionTrackingMigrated` keys off `submission_mode` presence only, so `grading` plays no part in the signal and can be dropped without touching the migration semantics. Leftover from #574.

The migration writes `submission_mode` and nothing else. `grading` stays absent, matching every other writer and the schema's writer convention.

### Steps to reproduce

1. In a classroom's `assignments.json`, have an entry without `submission_mode` and without `grading` (any pre-1.28 file).
2. Open the classroom in the web app; the migration banner appears. Click migrate.
3. Inspect the resulting commit on the config repo: the entry now carries `"grading": { "mode": "auto" }`.
4. Open that assignment in the web app's edit form and click Save without changing anything: a new "Edit assignment" commit lands whose only diff is the `grading` block disappearing.

### Version and environment

Web app at `main` (9489db85, 2026-08-21). Browser-independent — it is in the domain layer.

### Relevant logs or output

```shell
# migration commit diff (abridged)
   "slug": "hw1",
+  "submission_mode": "tag",
+  "grading": {
+    "mode": "auto"
+  },
```

### Before submitting

graphics80:
Closes #680

## What

`normalizeEntryForMigration` stamped an explicit `grading: { mode: "auto" }` onto every legacy entry that had no `grading` block. This PR drops that write; the migration now sets `submission_mode` and nothing else.

## Why

Every other writer omits the block for plain auto — the CLI (`omitempty`, no `--grading` flag) and the web's own `buildAssignmentEntry` ("mirroring the CLI's omitempty"). The schema reads absent as `auto` and documents "writers omit the default" for every defaulted field; `submission_mode` is the one deliberate exception because its presence is the migration signal, and `grading` has no such exemption.

Net effect of the old behavior: the first web edit after migration dropped the block again (shape flip without teacher intent), and the migration commit carried a `grading` diff nobody asked for. `isSubmissionTrackingMigrated` keys off `submission_mode` only, so `grading` never mattered for the signal.

## Scope

Deliberately narrow — three files, no schema change:

- `createEdit.ts`: remove the `grading` write + update the two comments.
- `assignments.test.ts`: the migration test now asserts `grading` stays absent (the "preserves existing grading verbatim" test is unchanged and still covers the manual case).
- `useMigrateClassroomAssignments.ts`: comment.

Independent of #654 / the `tag` → `every-push` question; touches only the `grading` lines, so the two can land in either order.

## Verification

- `vitest run src/domain/assignments.test.ts` — 182 passed
- `tsc -b`, `eslint`, `prettier --check` on the touched files — clean

### DIFF
--- a/web/src/domain/assignments.test.ts
+++ b/web/src/domain/assignments.test.ts
@@ -4093,7 +4093,7 @@ describe("migrateClassroomAssignments", () => {
     ...extra,
   })
 
-  it("writes explicit submission_mode + grading:auto onto every legacy entry in one commit", async () => {
+  it("writes explicit submission_mode onto every legacy entry in one commit, nothing else", async () => {
     const { client, committed } = makeMigrateClient([
       legacyEntry("hw1"),
       legacyEntry("hw2"),
@@ -4107,9 +4107,11 @@ describe("migrateClassroomAssignments", () => {
     expect(result.alreadyMigratedCount).toBe(0)
     const written = committed()!
     // Every entry now carries an explicit tag mode (preserving pre-1.28
-    // submit/*-release counting) and auto grading.
+    // submit/*-release counting). grading stays absent: the schema reads
+    // absent as auto and every other writer omits the block for plain auto,
+    // so stamping an explicit auto here would be the lone writer doing so.
     expect(written.match(/"submission_mode": "tag"/g)).toHaveLength(3)
-    expect(written.match(/"mode": "auto"/g)).toHaveLength(3)
+    expect(written).not.toContain(`"grading"`)
   })
 
   it("no-ops when every entry is already migrated (no commit)", async () => {
--- a/web/src/domain/assignments/createEdit.ts
+++ b/web/src/domain/assignments/createEdit.ts
@@ -1519,8 +1519,12 @@ function isSubmissionTrackingMigrated(entry: Assignment): boolean {
 }
 
 // Normalize an unmigrated entry into the new canonical shape by writing the
-// fields that pre-1.28 left absent: submission_mode = "tag" and, when grading
-// is absent, an explicit grading: { mode: "auto" } so the file self-documents.
+// one field that pre-1.28 left absent: submission_mode = "tag". Nothing else
+// is stamped on — in particular NOT an explicit grading: { mode: "auto" }:
+// the schema reads an absent grading as auto and every writer (CLI omitempty,
+// buildAssignmentEntry above) omits the block for plain auto, so writing it
+// here would only produce a shape no other path preserves (the next edit
+// would drop it again) and a noisy diff in the config repo.
 // "tag" (NOT "every-push") is what preserves pre-1.28 counting: before 1.28 a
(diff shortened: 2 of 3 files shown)


graphics80:
Closing — reopening properly after following the repo's contribution process.

PR #682, opened 2026-08-21
fix(web): stop the classroom migration from stamping grading: auto

### LINKED ISSUE #680: web: classroom migration stamps an explicit grading: { mode: "auto" } that every other writer omits

graphics80:
## Summary

The classroom migration (`normalizeEntryForMigration`) stamped an explicit `grading: { mode: "auto" }` onto every legacy entry without a `grading` block. No other writer does: the CLI has no `--grading` flag and omits the pointer (`omitempty`), and the web's own `buildAssignmentEntry` omits the block for plain auto to mirror it. The schema reads absent as `auto` and documents "writers omit the default" for every defaulted field; only `submission_mode` is exempted, as the migration signal. The explicit auto was therefore a shape no other path preserves (the next edit dropped it again) plus an unasked-for diff in the config repo's history.

This PR drops the `grading` write; the migration now sets `submission_mode` and nothing else. Four files (the function, its test, and two stale comments), no schema change. Independent of #654 — touches only the `grading` lines.

Files migrated by web-v1.28..v1.31 already carry the explicit block; it stays valid (the schema allows it, every reader uses `grading?.mode ?? "auto"`, and `isSubmissionTrackingMigrated` keys off `submission_mode` only, so nothing re-migrates). The app drops it on the next edit of that assignment; the CLI carries an existing block forward as-is.

Closes #680

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

  - CLI (Go): `go build ./... && go test ./... && golangci-lint run` in the module dir (`cli/gh-teacher`, `cli/gh-student`, or `cli/shared`) — not touched
  - Web: `cd web && npm run check` — clean (tsc, eslint, prettier, arch:validate, 4101 tests)
  - Skeleton scripts (Python): `python3 -m pytest cli/gh-teacher/skeleton_tests -q` — not touched
- [x] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass. — no contract change
- [x] If I added or changed a CLI command or flag, I documented it in the [wiki](https://github.com/foundation50/classroom50/wiki) (not in a README). — n/a

### DIFF
--- a/web/src/components/modals/MigrateSubmissionTrackingModal.tsx
+++ b/web/src/components/modals/MigrateSubmissionTrackingModal.tsx
@@ -20,10 +20,10 @@ type Phase = "idle" | "working" | "complete" | "error"
 // Greppable tag: MIGRATION(v1.28).
 // Explicit, teacher-triggered migration of a classroom's assignments.json to
 // the new submission-tracking semantics: one confirm writes an explicit
-// submission_mode (+ grading:auto) onto every legacy assignment, opting the
-// submissions-page detection overlay in without changing any grade. Nothing
-// runs until the teacher clicks Migrate. A content normalization within v1 —
-// no schema version bump.
+// submission_mode onto every legacy assignment, opting the submissions-page
+// detection overlay in without changing any grade. Nothing runs until the
+// teacher clicks Migrate. A content normalization within v1 — no schema
+// version bump.
 export function MigrateSubmissionTrackingModal({
   open,
   onClose,
--- a/web/src/domain/assignments.test.ts
+++ b/web/src/domain/assignments.test.ts
@@ -4093,12 +4093,9 @@ describe("migrateClassroomAssignments", () => {
     ...extra,
   })
 
-  it("writes explicit submission_mode + grading:auto onto every legacy entry in one commit", async () => {
-    const { client, committed } = makeMigrateClient([
-      legacyEntry("hw1"),
-      legacyEntry("hw2"),
-      legacyEntry("hw3"),
-    ])
+  it("writes explicit submission_mode onto every legacy entry in one commit, nothing else", async () => {
+    const legacy = [legacyEntry("hw1"), legacyEntry("hw2"), legacyEntry("hw3")]
+    const { client, committed } = makeMigrateClient(legacy)
     const result = await migrateClassroomAssignments(client, {
       org: ORG,
       classroom: CLASSROOM,
@@ -4106,10 +4103,12 @@ describe("migrateClassroomAssignments", () => {
     expect(result.migratedCount).toBe(3)
     expect(result.alreadyMigratedCount).toBe(0)
     const written = committed()!
-    // Every entry now carries an explicit tag mode (preserving pre-1.28
-    // submit/*-release counting) and auto grading.
-    expect(written.match(/"submission_mode": "tag"/g)).toHaveLength(3)
-    expect(written.match(/"mode": "auto"/g)).toHaveLength(3)
+    // Each entry gains exactly submission_mode: "tag" (preserving pre-1.28
+    // submit/*-release counting); grading stays absent, as every other writer
+    // leaves it (see normalizeEntryForMigration).
+    expect(JSON.parse(written).assignments).toEqual(
+      legacy.map((e) => ({ ...e, submission_mode: "tag" })),
+    )
   })
 
   it("no-ops when every entry is already migrated (no commit)", async () => {
--- a/web/src/domain/assignments/createEdit.ts
+++ b/web/src/domain/assignments/createEdit.ts
@@ -1519,8 +1519,12 @@ function isSubmissionTrackingMigrated(entry: Assignment): boolean {
 }
 
 // Normalize an unmigrated entry into the new canonical shape by writing the
-// fields that pre-1.28 left absent: submission_mode = "tag" and, when grading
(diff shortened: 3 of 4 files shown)


rongxin-liu:
Superseded by #683, which deletes `normalizeEntryForMigration` entirely. With no writer left, the stamp this PR removes cannot come back, and #680's expected outcome ("the migration writes `submission_mode` and nothing else") is satisfied. #683 now closes both #654 and #680.

Closing rather than merging, since merging first would only create a conflict on lines #683 deletes. The diagnosis here was correct, and it is what surfaced the shared root cause. Thanks for filing it.

Two corrections worth recording, because both bear on #683's reasoning.

**1. The schema citation is inverted.** This PR's summary says the schema "documents 'writers omit the default' for every defaulted field; only `submission_mode` is exempted, as the migration signal." The schema says the opposite for that field:

> writers omit `"every-push"` (the CLI normalizes it away), so readers must treat an absent field as `"every-push"`

That exemption was a web-layer invention. The phrase "migration signal" appears nowhere in `schemas/` or `cli/`, which is exactly the defect #654 reported. Worth flagging because PRs here are squash-merged, so the body becomes the permanent commit record.

**2. The `grading` convention is upheld independently.** `buildAssignmentEntry` still collapses `grading: auto` to omitted, pinned by "collapses grading:auto to omitted on an auto edit" in `assignments.test.ts`. So the convention this PR defended stays under test after #683's deletion.

For completeness: #683 also collapses an explicit `every-push` in `buildAssignmentEntry`. Presence was only ever load-bearing as the migration signal, so once that was gone an explicit `every-push` became a shape no other writer produced, and a form save on a CLI-created assignment added a key the teacher never touched.

graphics80:
thank you :-)

PR #688, opened 2026-08-21
fix(web): stop counting the tool's own commits as submissions

### LINKED ISSUE #687: submissions page counts the tool's own commits as student submissions
### Which part is affected?

Web app (classroom50.org)

### What happened?

Classroom 50 writes two commits into a student's repository on its own, both onto the repository's **default branch**, and the submissions page counts both as student submissions:

1. **The Feedback PR opener, at accept time.** When `feedback_pr` is on, the PR is created with `base: feedback, head: <default branch>`. GitHub refuses a PR with no commits between base and head — the normal case at accept — so the app lands one empty commit and retries. That commit fast-forwards the **default branch**, not the feedback branch.
   `[Classroom 50] Open Feedback PR (gh student accept)` + `[skip ci]`
2. **The submission-mode retrofit.** *Update autograding triggers* (or `gh teacher assignment submission-mode`) rewrites each accepted repository's workflow with one commit per repository.
   `[Classroom 50] Update autograder trigger to <every-push|tag> (submission-mode)` + `[skip ci]`

On an **every-push** assignment, `detectBranchSubmissions` excludes only the baseline (the accept commit that created `.classroom50.yaml`), so both of the above are counted. Consequences:

- Every accepted repository shows **1 submission before the student has pushed anything**, with a "last submitted" time — the accept timestamp. On a `no_autograder` / `empty_repo` assignment, where detection is the only source, that is the entire displayed count.
- Every trigger retrofit adds another one to every accepted repository.
- Because detection can only reveal *more* submissions than the collected snapshot holds (max wins in `mergeDetectedSubmissions`), the affected rows are also flagged as having an out-of-date count — pointing at work that can never be graded: both commits carry `[skip ci]`, and `runner.py` skips the retrofit commit even without it (`is_shim_update_commit`).
- The student's own submission list has the same off-by-N: `useGetMyPushSubmissions` re-derives the same baseline-only filter.

This was invisible until #683: the detection overlay used to be gated on the classroom having been "migrated", so only classrooms that had clicked the old banner saw these counts. With the gate gone, every classroom does.

Found by code inspection while working through the #654 follow-ups, not from a support report.

Second, smaller mismatch in the same area: the contract says a **push** is the counting unit (`submission_mode` in `schemas/assignments-v1.schema.json`: *"every default-branch push (minus the baseline commit) counts as a submission"*, and the UI reads *"N pushes to the default branch"*), while `detectBranchSubmissions` has always counted one submission per **commit**. A push of three commits is one graded run and one Release, but three counted submissions.

A freshly accepted repository counts 0 submissions until the student pushes, and a teacher-side trigger update doesn't change any count. Commits the tool authors for its own bookkeeping — which no autograder run can ever follow — should not appear as student work in either the teacher's or the student's view. And the counting unit should read the same in the schema, its mirrors and the UI.

### Steps to reproduce

1. Create an assignment with **Submission type = Every push to the default branch** and **Feedback pull request** enabled.
2. Have a student accept it, and push nothing.
3. Open the assignment's submissions page: the student is counted with **1 submission**, "last submitted" showing the accept time, although the repository contains only the accept and Feedback-PR commits and Actions shows no graded run.
4. Optional, for the second commit: switch **Submission type** to *A tagged commit*, run **Update autograding triggers**, switch back to *Every push* and run it again. The count rises by 2 more, with no run and no Release behind either.

### Version and environment

Web app at `main` (7593797d), also live on classroom50.org. Not browser-specific — the miscount is in the client-side detection derivation (`web/src/domain/assignments/submissionDetection.ts`).

### Relevant logs or output

_No response_

graphics80:
## Summary

Classroom 50 writes two commits of its own onto a student repository's **default branch**, and branch-mode detection counted both as student submissions:

- the empty commit that opens the Feedback PR at accept time (`pushEmptyCommit` fast-forwards `head`, which is the default branch — the zero-diff 422 it answers is the normal case at accept), so **every accepted repository showed 1 submission before the student pushed anything**, timed at the accept;
- the submission-mode shim retrofit, which adds one more to every accepted repository each time a teacher runs *Update autograding triggers*.

Neither can ever be graded — both carry `[skip ci]`, and the runner skips the retrofit commit even without it (`is_shim_update_commit`) — yet because detection can only reveal *more* submissions than the collected snapshot holds, the affected rows were additionally flagged as having an out-of-date count. On a `no_autograder` / `empty_repo` assignment, where detection is the only source, that phantom was the whole count.

`submissionCommits(commits, baselineSha)` is now the one place that decides which default-branch commits are submissions, and both readers go through it — the teacher's fan-out and the student's own view, which previously spelled the baseline-only filter out separately and could drift apart.

The excluded set is built from the writers' own message constants rather than from copies, and it is an explicit enumeration, not "anything carrying `[skip ci]`": a student can write that marker too, and their push is still a submission, it just wasn't graded. The student's own `[Classroom 50] Submit <slug>` commit is therefore untouched, and anything unrecognized still counts — an unknown commit is a submission, never a silent skip. Tests pin the exclusion against both writers, so a writer that drops its marker fails the suite instead of silently inflating counts again.

Those two message constants move to `web/src/util/commit.ts`, next to `COMMIT_PREFIX`, for the same reason Go keeps them in `cli/shared/contract` and not in the command packages: the writers and the submissions page both need them, and the detection module should not import a writer to get a string.

This was invisible until #683 — the overlay used to be gated on the classroom having been "migrated", so only classrooms that had clicked the old banner saw these counts. Removing the gate widened the reach, which is what surfaced it.

### Count the baseline as a cut point, not a single SHA

A follow-up fixed a second undercount source found while validating against live repos. `submissionCommits` excluded only the *single* commit whose SHA equals the baseline, but on a **templated** assignment the accept commit (the `.classroom50.yaml` marker, which is the baseline) sits *on top of* the template's generated `Initial commit`. That template commit is **older** than the baseline, so a plain `sha !== baselineSha` filter left it counting — a freshly-accepted, never-pushed templated repo showed **1** instead of **0**.

Commits arrive newest-first, so `submissionCommits` now treats the baseline as a **cut point**: it keeps only the commits before the baseline's position (the genuinely newer student work), drops the baseline and every ancestor after it (the whole accept-time template prefix), and then removes the tool's bookkeeping subjects. A regression test pins the real-world history (`Initial commit` -> accept/baseline -> Feedback-PR opener, no student push -> **0**).

Note this only corrects the **detection** overlay. A repo accepted with an *older skeleton* (before the `--detect-acceptance` gate in `autograde-runner.yaml`) may still carry a real `submit/*` Release on its accept commit, which the collected snapshot counts as one graded submission independently of detection; the current skeleton already skips grading the accept commit, so newly accepted repos don't produce that phantom.

### The counting unit

While here, the contract said one thing and the code did another: `submission_mode` documented **pushes** as the counting unit while `detectBranchSubmissions` has always counted **commits**, and the UI read "N pushes to the default branch" for what were commits. A push of several commits is one graded run and one Release but several counted submissions, so the schema, the TypeScript mirror, the counting copy in `en.json` and the wiki now say *commit past the baseline*. The mode's own **label** still says push, because the grading trigger really is per push — the schema now states that distinction instead of leaving it implicit.

Grouping commits into pushes via `GET /actions/runs?event=push` was considered as the alternative and rejected: one more paginated read per repository in a fan-out whose per-repo reads are already unbounded, and pushes that never ran (Actions disabled, `no_autograder`) would have no run to group by.

The Go and Python contract copies describe when the shim *fires*, not what the page counts, so their wording stands; the one Go comment that did name the counting unit is updated. `collect_scores.py` counts `submit/*` releases only and has no commit-counting twin. `runner.py`'s `is_shim_update_commit` docstring now records that the web answers the same question from the commit **subject** (a fixed set of tool subjects — not the `[skip ci]` body or touched paths, since the list-commits API carries no per-commit paths), and that the two rules therefore differ at the edge.

### Review follow-ups

A code review of the original commit surfaced a few loose ends, addressed here:

- The schema/TS text claiming the submissions page "counts default-branch commits" for `empty_repo`/`no_autograder` is narrowed to the **student view**: the teacher page gates live detection on `liveCapable = isOwner && !skipsGrading`, which is off for those shapes, so it renders the collected snapshot instead.
- The student reader `useGetMyPushSubmissions` gains a hook-level exclusion test (its sibling `useDetectedSubmissions` already had one), plus an exact-subject-collision test documenting the accepted subject-only trade-off.
- The `en.json` help, `wiki/Glossary.md`, `wiki/CLI-Student-Guide.md`, and `wiki/Web-Teacher-Guide.md` are reconciled to distinguish per-push grading from the per-commit count and to name the excluded tool commits, and the schema nit ("the two commits the tool authors itself" -> "authored by the tool") is applied.
- Because the two commit-message constants moved into `web/src/util/commit.ts`, the Go contract's hand-mirror comments (`cli/shared/contract/contract.go`) are repointed there — those comments are the only cross-language link for these byte-identical strings, so they move with the code. `contract_test.go` still pins the parity.
- A simplification pass extracted the branch-mode read router both hook tests hand-rolled into `@/test/branchDetectionClient`. The student-hook copy resolved `null` for a not-accepted repo, bypassing `getRepo`'s `tolerateGitHubError`; the shared router rejects 404 like the real client, so both tests now exercise that path. Same pass types `shimUpdateCommitMessage` with `SubmissionMode` rather than respelling the union, and trims the comment blocks to the load-bearing invariants.

Closes #687

## Type of change

- [x] Bug fix


rongxin-liu (approved):
Thanks!

PR #694, opened 2026-08-22
fix: show submissions for assignments that skip autograding

### LINKED ISSUE #659: Can't see students submissions
### Which part is affected?

Web app (classroom50.org)

### What happened?

I tried creating various assignments (with and without templates, with push on the default branch and with a tag, with and without autograding, manual grading) and the assignments web page always show 0/1 submissions. 

Image

When going in the specific assignment details screen, students don't have anything in "Last submitted" even if there are commits and tagged releases. 

Image

What is strange is that the "Manage submissions" button show to commit/push date correctly and cloning with `gh teacher` is working fine too.

I expect to be able to monitor easily students that have accepted (or not) and students that have pushed.

### Steps to reproduce

Just using the web ui as documented on https://github.com/foundation50/classroom50/wiki/Web-Teacher-Guide

### Version and environment

Windows 11
gh version 2.97.0 (2026-07-31)

### Relevant logs or output

```shell

```

### Before submitting

rongxin-liu:
## Summary

Teachers reported that assignments with autograding disabled always showed `0/1 submissions` and a blank "Last submitted", even after students had committed, pushed, and tagged. `gh teacher` cloning and "Manage submissions" worked fine, which made the web app look simply wrong.

Two independent causes, both fixed here:

**Nothing was ever collected.** `collect_scores.py` skipped these assignments outright (`no_autograder assignment — autograding is disabled; skipping collection`), so their `scores.json` bucket stayed permanently empty. It only ingests `submit/*` releases, which these repos never publish. Collection now also *detects* submissions from repo state — default-branch commits past the accept baseline, or submission tags — and records them as presence and count. No score is ever synthesized: records go in a new optional `detected` list, never in `entries`. `skips_grading` still governs every grading path, and a bare `empty_repo` assignment stays skipped since it has no submission definition to detect against.

**The detail page had the data and hid it.** The commit/tag detection overlay was gated behind `liveCapable`, which is false for exactly these assignments. Split into `liveCapable` (submit/* releases) and `detectionCapable` (raw repo state), so a no-autograder assignment now shows its submissions and a real "Last submitted".

The assignments list renders the same `N / M` and progress bar as an autograded row, from the same single `scores.json` read — so both row types are consistent, and one derivation of "submitted" cannot disagree with another. An absent `detected` list means no collect has walked the bucket yet, which is *not* "nobody submitted", so that case says so rather than showing a misleading `0 / M`.

Along the way: the freshness line and "Refresh submissions" button were hidden for every assignment that skipped grading. Now that these are genuinely collected, both are restored (still hidden for `empty_repo`).

Closes #659

Also expected to resolve the same class of report in [discussion #677](https://github.com/foundation50/classroom50/discussions/677). Unrelated to [#689](https://github.com/foundation50/classroom50/discussions/689) (collect permissions) and [#675](https://github.com/foundation50/classroom50/discussions/675) (stuck Actions runs), which are separate collect-infrastructure issues.

### Why collection rather than reading repos from the browser

Deriving the count client-side was the obvious alternative and does not work at scale: branch-mode detection costs at least 3 API calls per repo (two of them fully paginated), throttled to 8 concurrent reads. A 10-assignment, 200-student class is roughly 2,000 repos and 6,000 calls on a single page load — past the 5,000/hour REST cap, and minutes of fetching for a screen teachers glance at. That is why the detail page is deliberately page-scoped. Client-side detection is also owner-only, so a TA or HTA would never see the bar. Collecting once in the scheduled Action costs the browser nothing, works for every role, and keeps one source of truth.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

  - CLI (Go): `go build ./...`, `go test ./...`, `golangci-lint run` (0 issues), `golangci-lint fmt --diff` (clean)
  - Web: `cd web && npm run check` — 4,248 tests, 0 lint errors, prettier + `arch:validate` clean
  - Skeleton scripts (Python): `pytest cli/gh-teacher/skeleton_tests -q` (855) and `autograders_tests -q` (388)
- [x] If I added or changed a CLI command or flag, I documented it in the wiki (not in a README).

## Contract change

`schemas/scores-v1.schema.json` gains an **optional** `detected` array on `assignmentBucket` (`detectedRecord`: `owner` + `count` required; optional `kind` / `latest_datetime` / `late` / `member_usernames`). Additive and optional, so older readers and writers are unaffected.

Deliberately *not* reused `submissionRecord`: it requires `score`, `max-score`, `tests`, `release`, `review`, and a `^submit/` submission. Writing a detected commit there would fabricate a grade and corrupt every grade consumer (stats, CSV export, `ScoreBadge`, passing/threshold filters). Mirrors updated in the same change — Python writer, TypeScript reader (`useGetScores`), and the Go reader (`internal/scores`, read-only on this file), plus 15 schema parity tests.

## Review notes

Two review passes ran over this branch and caught five real defects, all fixed here. Worth calling out for reviewers:

- The tag matcher initially used `fnmatch`, which disagreed with `cli/shared/testdata/submission_tag_match_cases.json` on **13 of 57 cases** (`*` crossed `/`, `?` acted as a wildcard, no stacked-quantifier guard). Detection would have claimed a different tag set than the shim that grades. Replaced with the mirrored implementation and pinned to the shared fixture like the Go, web, and regrade copies.
- Tag mode no longer records `latest_datetime` or `late`: that instant comes from the student-authored `submit/<ts>` tag **name**, which can be backdated to dodge a late flag. The web refuses tag times for lateness for the same reason. The count still stands — tag existence is not forgeable.
- Detected records **merge per owner** rather than replacing the list, so a repo whose read failed keeps its prior record instead of a transient 500 silently deleting a recorded submitter.
- Metrics is now gated on the overlay union: `computeStats` skips `pending` rows, so it would have reported 0 submitted beside a table listing detected submitters.

## Test plan

- [x] Unit tests: collector detection (branch + tag mode, late marking, non-submitter omission, unreadable-repo skip, record persistence), shared-fixture matcher parity, `normalizeScores` mirror, `AssignmentsTable` cell states, `computeStats` all-pending invariant.
- [x] New behavior tests verified red-before-green, including the destructive-replacement case.
- [ ] **Manual verification still needed** (needs a real classroom + a collect run): create a no-autograder assignment, have a student push, run Collect Scores, and confirm the list shows `1 / N` with the bar and the detail page shows "Last submitted". Please also sanity-check the tag-mode path against a real `submit/*` tag.

## Post-Deploy Monitoring & Validation

- **Logs to watch:** the Collect Scores workflow for `no_autograder assignment — autograding is disabled; detected N submitter(s) from repo state`. The old `skipping collection` line should now appear only for `empty_repo`. Watch for a spike in `submission detection failed: HTTP ...` warnings, which would indicate token scope or rate-limit trouble on the new repo reads.
- **Metrics:** Collect Scores run duration and failure rate. Detection adds up to 3 API calls per no-autograder repo, so runtime grows with the number of such assignments; a large org adopting them heavily is the case to watch.
- **Healthy signals:** collect finishes green; `detected` appears in `<classroom>/scores.json`; the assignments list shows a non-zero bar for assignments with real pushes.
- **Failure signals / rollback:** collect runs timing out or hitting secondary rate limits, or `detected` counts disagreeing with the submissions page. Detection is additive and grade-free, so rolling back the CLI restores the previous behavior with no data migration — a stale `detected` list is ignored by older web builds.
- **Validation window:** the first nightly collect after release, plus one manual "Sync now" on a no-autograder assignment.

> [!IMPORTANT]
> Web and CLI must ship together — the bar only populates after a collect runs with the new CLI. release-please's `linked-versions` already enforces the lockstep.


ncarvajalc:
@rongxin-liu was looking at this fix and there is still a bug, even in preview. In which a TA cannot see the submissions, but the org owner can:

This is my teacher view as codeowner:
image

And this is my TAs view:
image

Right now my TAs (or head TAs) are not able to see the submissions in any assignment.

I document this behavior in https://github.com/foundation50/classroom50/discussions/677#discussioncomment-18141570 too

PR #720, opened 2026-08-24
feat(web): collect scores for a whole classroom

### LINKED ISSUE #719: Collect scores for every assignment in a classroom from the UI
## Which part would this affect?

`Web app (classroom50.org)`

## Problem or current behavior

The web app can only trigger score collection for **one assignment at a time**. The
"Collect now" item lives in the submissions-page actions menu
(`web/src/pages/submissions/SubmissionsActionsMenu.tsx`), and the page dispatches
`collect-scores.yaml` scoped to that single slug
(`web/src/pages/SubmissionsPage.tsx`, `useTriggerScoreCollection`).

A teacher who wants a fresh gradebook for the whole classroom — end of week, before a
grade export, after a batch of late submissions — has to open every assignment's
submissions page and press "Collect now" once per assignment, each dispatch queueing
behind the previous one on the same `collect-${classroom}` concurrency group.

The capability already exists everywhere except the UI:

- `collect-scores.yaml` declares `assignment` as optional — "leave blank to collect
  the whole classroom".
- `triggerScoreCollection(client, org, scope)` already types `assignment` as optional
  (`web/src/github-core/mutations/workflowDispatch.ts`).

Only `useTriggerScoreCollection` narrows it back to required
(`CollectScoresScope = { classroom: string; assignment: string }`), and no page other
than the submissions page calls it. So today the sole way to run a classroom-wide
collection is the GitHub Actions tab of `<org>/classroom50`: *Collect Scores → Run
workflow → fill in `classroom`, leave `assignment` blank* — which means leaving the
app and knowing the config repo exists.

## Proposed solution

A "Collect all" action on the classroom's assignments page (`/$org/$classroom/assignments`),
next to "New assignment" in the toolbar, that dispatches `collect-scores.yaml` with only
the `classroom` input — one run that walks every assignment in the classroom.

Details that matter:

- Available to any staff viewer (a TA can already collect per assignment; only
  authoring is author-gated).
- Hidden on an archived classroom and when the classroom has no assignments yet.
- Disabled with an explanatory title while the roster is empty (mirrors
  `submissions.collect.titleEmptyRoster`).
- In-flight state on the button, outcome as a toast; the run also shows up in the
  existing Actions activity banner, like the per-assignment dispatch.
- On success, invalidate the classroom's `scores.json` query so the submission counts
  in the table re-derive.
- No workflow-version guard needed: `classroom` is the long-standing input, so the
  `CollectInputsUnsupportedError` (422 "unexpected inputs") case is specific to the
  newer per-assignment scope and cannot fire here.

## Alternatives considered

- **Run the workflow from the Actions tab** — works today, but requires leaving the app
  and knowing the config repo layout; not something to put in front of a teacher.
- **Org-wide "collect everything" button** — the dispatch supports it (omit both inputs),
  but it crosses classroom boundaries and would let one teacher rebuild another's
  gradebook; classroom scope is the natural unit of the permission model.
- **Fan out one per-assignment dispatch per assignment from the client** — same result,
  N runs serialized on the same concurrency group, N times the API traffic, and a
  partial failure leaves the gradebook half-refreshed.
- **Rely on the scheduled collection** — fine for the steady state, but the whole point
  of the manual action is not waiting for it.

## Additional context

Relevant code:

- `cli/gh-teacher/skeleton/dotgithub/workflows/collect-scores.yaml` — inputs and the
  `collect-${{ inputs.classroom || 'all' }}` concurrency group.
- `web/src/github-core/mutations/workflowDispatch.ts` — `triggerScoreCollection`, scope
  already optional.
- `web/src/hooks/useTriggerScoreCollection.ts` — the hook that narrows it.
- `web/src/pages/AssignmentsPage.tsx` — where the action would live.

## Before submitting

graphics80:
## Summary

The web app could only collect scores one assignment at a time — a teacher refreshing a classroom's gradebook had to walk every submissions page and press "Collect now" on each, or run `collect-scores.yaml` by hand from the Actions tab.

This adds a **Collect all** action to the classroom's assignments toolbar. It dispatches the same workflow with only the `classroom` input, so one run walks every assignment in the classroom.

The dispatch layer already supported it (`triggerScoreCollection`'s `scope.assignment` was already optional, and the workflow documents a blank `assignment` as "collect the whole classroom") — only `useTriggerScoreCollection` narrowed `assignment` back to required, and no page but the submissions page called it.

Behavior:

- The action leads the toolbar, mirroring the submissions page's layout: the collect affordance sits left (where DataFreshness leads there), and search + filters + sort + the New assignment button sit right, so the two toolbars read the same way.
- The button is paired with a passive freshness line ("Submission data synced x ago", reusing the shared `submissions.freshness.*` strings), so the context is implicit: the table's submission counts come from the `scores.json` snapshot this button rebuilds. The timestamp is the newest per-bucket `collected_at` stamp — free, since the table already has `scores.json` in the query cache — and a stamped file never borrows the org-wide run timestamp (that run may have swept a different classroom); only a wholly unstamped pre-stamp-collector file falls back to it, the same precedence `effectiveCollectedAt` applies per bucket.
- Clicking confirms before dispatching: a sweep takes longer for a large classroom and uses more GitHub Actions minutes than a single-assignment collect, so the modal says so and the dispatch only fires from its confirm action.
- Visible to any staff viewer — a TA can already collect per assignment; only authoring is author-gated. Hidden on an archived classroom and while the classroom has no assignments; disabled with an explanatory title while the roster is empty.
- A sweep's execution is tracked against the workflow's own 30-minute job cap (`collect-scores.yaml`, `timeout-minutes: 30`) instead of the 10-minute default tuned for a single assignment. That cap measures execution only, and the sweep shares the `collect-${classroom}` concurrency group with every per-assignment collect in the classroom, so `useGitHubOperation` now waits out the Actions queue on its own window before the execution window starts counting (anchored on `run_started_at`) — previously a healthy queued run could trip the client timeout, which drops the tracked dispatch and re-enables the button mid-run, inviting a duplicate queued behind it.
- The outcome is reported by the existing app-wide Actions banner, which the hook already registers every dispatch with and which survives navigating away mid-sweep. The one case it cannot cover is a dispatch that never lands — registration rides the mutation's success — so a rejected POST (no config-repo write, no `classroom50` repo) gets an error toast, reusing the submissions page's wording for the same failure. The hook now attributes a failure to the dispatch or the run (`failure: "dispatch" | "run"`), and the toast fires only for the dispatch side — a run that concluded non-success is already the banner's to report.
- A finished run invalidates this classroom's `scores.json` and the last-collect-run stamp, so the table's submission counts and the freshness line re-derive. A client-side `timeout` invalidates too: the poll giving up says nothing about the run, and leaving the page on stale counts is the worse default.
- No workflow-version guard: `classroom` has been a dispatch input since #58, while `assignment` arrived in #593 — so the `CollectInputsUnsupportedError` (422 "unexpected inputs") case is specific to the per-assignment scope and cannot fire here.

The wiki pages that describe score collection now cover the classroom-wide collect: a new "Collect the whole classroom" section in the Web Teacher Guide, and updated mentions in GitHub Integration (the dispatch-scope examples), Autograding Basics, How Classroom 50 Works, and Managing Actions Cost (what a sweep bills). They publish with the merge, since the wiki syncs from `wiki/` on `main`.

Closes #719

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

  - Web: `cd web && npm run check` — passes (tsc, eslint, prettier, arch:validate, 352 test files / 4416 tests).
- [x] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass. — n/a, no contract change: the workflow inputs and the `scores.json` shape are untouched.
- [x] If I added or changed a CLI command or flag, I documented it in the wiki. — no CLI change, but the wiki pages describing score collection are updated for the classroom-wide collect (Web Teacher Guide, GitHub Integration, Autograding Basics, How Classroom 50 Works, Managing Actions Cost).

## Files

| File | Change |
| --- | --- |
| `web/src/hooks/useGitHubOperation.ts` | queue-aware deadline (`queueTimeoutMs`, anchored on `run_started_at` once the run leaves the queue) and failure attribution (`failure: "dispatch" \| "run"`) |
| `web/src/hooks/useGitHubOperation.test.tsx` | new — queue-wait windows, skewed start stamps, and failure attribution, on fake timers |
| `web/src/hooks/useTriggerScoreCollection.ts` | `CollectScoresScope.assignment` optional; scope suffix keeps a sweep and an assignment collect on distinct storage/query keys; 30-minute execution window for a sweep; forwards `failure` |
| `web/src/hooks/useTriggerScoreCollection.test.tsx` | new — scope keying and the timing window, without mocking the hook under test |
| `web/src/pages/assignments/ClassroomCollectButton.tsx` | new — freshness line + button + confirm modal, cache invalidation, dispatch-failure toast |
| `web/src/pages/assignments/ClassroomCollectButton.test.tsx` | new — confirm-then-dispatch flow, cancel, empty-roster gate, freshness-line derivation, invalidation on completed/timeout, dispatch-vs-run failure toast |
| `web/src/pages/assignments/AssignmentsToolbar.tsx` | `leading` slot for the collect action; search + filters move into the trailing group, mirroring `SubmissionsControls` |
| `web/src/pages/AssignmentsPage.tsx` | passes the collect action as `leading` and the primary action as `trailing` |
| `web/src/pages/AssignmentsPage.test.tsx` | page-level gating of the collect action: leads the toolbar for staff with assignments, absent when archived or the list is empty |
| `web/src/locales/en.json` | `assignments.collect.label`, `.title`, and the confirm-modal strings; the freshness and status strings are reused from `submissions.freshness.*` / `submissions.collect.*` rather than duplicated across 18 language packs |
| `web/src/github-core/mutations/workflowDispatch.test.ts` | a classroom-only dispatch must omit the `assignment` key entirely, not send it blank |
| `wiki/` (5 pages) | Web Teacher Guide gains a "Collect the whole classroom" section; GitHub Integration, Autograding Basics, How Classroom 50 Works, and Managing Actions Cost mention the classroom-wide collect where they describe collection |

## Not in this PR

- **Org-wide collect.** The dispatch supports omitting both inputs, but it crosses classroom boundaries and wants its own permission discussion.
- **Dispatch-to-run binding.** Run discovery identifies our run as the oldest one past the pre-dispatch baseline, and the sweep and per-assignment trackers share that lookup. Two dispatches snapshotting the same baseline within the seconds before the first run appears can bind to the same run — a pre-existing property of every dispatch surface (the dispatch API returns no run id to correlate on), with mild consequences (the concurrency group serializes the runs, and completion invalidates the same caches). Correlating via a dispatch marker input is its own change.
- **Classroom-wide staleness.** The submissions page's red "Sync now" (push after last collect) needs every assignment's repos' `pushed_at` correlated against per-bucket stamps; the freshness line here stays passive.
- **One presenter for `OperationPhase`.** The submissions status strip, the activity banner, and this button are three hand-written mappings of the same five states; unifying them is worth doing once a third collect surface lands, not here.
- **The second poller.** `useGitHubOperation` polls the run while the activity banner polls the same workflow's runs anyway — pre-existing for every dispatch site, not introduced here.
- **`OP_TTL_MS`.** The activity banner drops a tracker after 15 minutes, so the banner row for a very long sweep can disappear before the run ends. Changing that TTL affects every tracked operation, so it is left alone.

### DIFF
--- a/web/src/github-core/mutations/workflowDispatch.test.ts
+++ b/web/src/github-core/mutations/workflowDispatch.test.ts
@@ -79,6 +79,24 @@ describe("triggerScoreCollection", () => {
     })
   })
 
+  // The classroom sweep: `assignment` must be absent, not empty — the workflow
+  // treats a present-but-blank input as "collect the whole classroom" too, but
+  // an org whose collect-scores.yaml predates the `assignment` input 422s on the
+  // key itself.
+  it("sends only the classroom input when the scope has no assignment", async () => {
+    const { client, request } = makeClient(() => ({}))
+
+    await triggerScoreCollection(client, "acme", { classroom: "cs50" })
+
+    const dispatchCall = request.mock.calls.find(([url]) =>
+      (url as string).endsWith("/dispatches"),
+    )
+    expect(
+      (dispatchCall?.[1] as { body: { inputs: Record<string, string> } }).body
+        .inputs,
+    ).toEqual({ classroom: "cs50" })
+  })
+
   it("maps a scoped 422 'unexpected inputs' to CollectInputsUnsupportedError", async () => {
     const { client } = makeClient(() => {
       throw unexpectedInputs422()
--- a/web/src/hooks/useGitHubOperation.test.tsx
+++ b/web/src/hooks/useGitHubOperation.test.tsx
@@ -1,0 +1,186 @@
+// @vitest-environment happy-dom
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
+import { act, cleanup, renderHook } from "@testing-library/react"
+import type { ReactNode } from "react"
+
+import { useGitHubOperation } from "./useGitHubOperation"
+import type { GitHubWorkflowRun } from "@/github-core/types"
+
+const MINUTE = 60 * 1000
+const TIMEOUT_MS = 10 * MINUTE
+const QUEUE_TIMEOUT_MS = 30 * MINUTE
+// The tracker never waits longer than the two windows back to back.
+const MAX_WAIT_MS = TIMEOUT_MS + QUEUE_TIMEOUT_MS
+const T0 = new Date("2026-08-24T12:00:00.000Z")
+const STORAGE_KEY = "test:operation"
+
+const at = (offsetMs: number) => new Date(T0.getTime() + offsetMs).toISOString()
+
+const makeRun = (over: Partial<GitHubWorkflowRun> = {}): GitHubWorkflowRun => ({
+  id: 2,
+  status: "queued",
+  conclusion: null,
+  created_at: at(0),
+  html_url: "https://github.com/acme/config/actions/runs/2",
+  event: "workflow_dispatch",
+  ...over,
+})
+
+// The run the poll currently sees, and what the dispatch does. Both are swapped
+// per test; the poll re-reads `currentRun` on every refetch, so a test can move
+// a run from queued to in_progress by reassigning it and advancing the clock.
+let currentRun: GitHubWorkflowRun | null = null
+let dispatchImpl: () => Promise<{ sinceRunId: number | null }> = async () => ({
+  sinceRunId: 1,
+})
+
+const mount = () => {
+  const client = new QueryClient({
+    defaultOptions: { queries: { retry: false, gcTime: 0 } },
+  })
+  const wrapper = ({ children }: { children: ReactNode }) => (
+    <QueryClientProvider client={client}>{children}</QueryClientProvider>
+  )
+  return renderHook(
+    () =>
+      useGitHubOperation({
+        storageKey: STORAGE_KEY,
+        queryKey: (sinceRunId) => ["operation", sinceRunId],
+        resetKey: "acme",
+        timeoutMs: TIMEOUT_MS,
+        queueTimeoutMs: QUEUE_TIMEOUT_MS,
+        // Poll on a flat one-minute cadence so the advance amounts below read
+        // as wall-clock, not as a backoff schedule.
+        intervalMs: MINUTE,
+        backoffAfterMs: MINUTE,
+        backoffIntervalMs: MINUTE,
+        dispatch: () => dispatchImpl(),
+        findRun: async () => currentRun,
+      }),
+    { wrapper },
+  )
+}
+
+const advance = async (ms: number) => {
+  await act(async () => {
+    await vi.advanceTimersByTimeAsync(ms)
+  })
+}
+
+// Dispatch and let the first poll land, so the hook is tracking a run.
+const start = async (view: ReturnType<typeof mount>) => {
+  await act(async () => {
+    view.result.current.trigger()
+  })
+  await advance(MINUTE)
+}
+
+beforeEach(() => {
+  vi.useFakeTimers()
+  vi.setSystemTime(T0)
+  sessionStorage.clear()
+  currentRun = null
+  dispatchImpl = async () => ({ sinceRunId: 1 })
+})
+
+afterEach(() => {
+  cleanup()
+  vi.useRealTimers()
+})
+
+// GitHub's own `timeout-minutes` caps execution only. A run held in an Actions
+// concurrency group is healthy, so the client must not expire it — doing so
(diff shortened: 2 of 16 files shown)


rongxin-liu (approved):
Thank you!

PR #735, opened 2026-08-25
fix(web): stop the assignments action column from stretching

graphics80:
## Summary

The assignments list's action column stretched: `.table` is `width: 100%` with
auto layout, so the browser spread the surplus width across every column —
including the one holding a fixed-width strip of icon buttons. The result was a
wide empty gap to the right of the last action (delete), and text columns that
were narrower than they needed to be.

`w-0` on the column's `<th>`/`<td>` makes the browser fall back to min-content
there and hand the slack to the text columns instead. While in the cell, the
buttons moved into a `flex items-center justify-end gap-1` wrapper — the same
recipe the submissions table's action cell already uses
(`web/src/pages/submissions/SubmissionsRows.tsx`) — and `py-2!` trims the cell's
own vertical padding, which `TableShell`'s `padded` had set to `py-4`; with a
32px button that cell, not the text cells, was driving the row height.

Visual only. No behaviour, markup semantics, or a11y affordances change: the
buttons keep their 32px hit targets, their labels, and their order.

**Review tip:** most of the diff is re-indentation from the new wrapper
`<div>`. `?w=1` shows the real change — 8 insertions, 2 deletions.

**Stacked on #732.** It rewrites the same action cell, so this branches off
`feat/copy-assignment-link` rather than `main`. Please merge after #732; I'll
rebase if that PR moves during review.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

  - Web: `cd web && npm run check` — 353 files, 4432 tests, all green
- [x] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass. — n/a, no contract touched
- [x] If I added or changed a CLI command or flag, I documented it in the [wiki](https://github.com/foundation50/classroom50/wiki) (not in a README). — n/a, no CLI change
- [x] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) (e.g., `feat(web): ...`, `fix(gh-teacher): ...`)


graphics80:
Branched off #732 to avoid a conflict, but a cross-fork PR can't use another fork's branch as its base — so this diff carried #732's commits along with it. Reopened cleanly off `main` as #736; whichever of the two lands second gets the rebase.

PR #736, opened 2026-08-25
fix(web): stop the assignments action column from stretching

graphics80:
## Summary

The assignments list's action column stretched: `.table` is `width: 100%` with
auto layout, so the browser spread the surplus width across every column —
including the one holding a fixed-width strip of icon buttons. The result was a
wide empty gap to the right of the last action (delete), and text columns
narrower than they needed to be.

`w-0` on that column's `<th>`/`<td>` makes the browser fall back to min-content
there and hand the slack to the text columns instead. While in the cell, the
buttons moved into a `flex items-center justify-end gap-1` wrapper — the recipe
the submissions table's action cell already uses
(`web/src/pages/submissions/SubmissionsRows.tsx`) — and `py-2!` trims the cell's
own vertical padding, which `TableShell`'s `padded` had set to `py-4`; with a
32px button, that cell rather than the text cells was driving the row height.

Visual only. No behaviour, markup semantics, or a11y affordances change: the
buttons keep their 32px hit targets, their labels, and their order.

**Review tip:** most of the diff is re-indentation from the new wrapper
`<div>`. `?w=1` shows the real change — 8 insertions, 2 deletions.

Heads-up: #732 rewrites the same action cell, so whichever of the two lands
second needs a rebase. Happy to take that side.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

  - Web: `cd web && npm run check` — 352 files, 4422 tests, all green
- [x] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass. — n/a, no contract touched
- [x] If I added or changed a CLI command or flag, I documented it in the [wiki](https://github.com/foundation50/classroom50/wiki) (not in a README). — n/a, no CLI change
- [x] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) (e.g., `feat(web): ...`, `fix(gh-teacher): ...`)

### DIFF
--- a/web/src/pages/assignments/AssignmentsTable.tsx
+++ b/web/src/pages/assignments/AssignmentsTable.tsx
@@ -474,7 +474,11 @@ const AssignmentsTable = ({
           />
           <th scope="col">{t("assignments.table.colAccepted")}</th>
           <th scope="col">{t("assignments.table.colSubmitted")}</th>
-          <th scope="col">
+          {/* w-0: auto table layout hands surplus width to every column,
+              which stretched this fixed-width button strip. Zero width makes
+              the browser fall back to min-content here and give the slack to
+              the text columns instead. */}
+          <th scope="col" className="w-0">
             <span className="sr-only">{t("assignments.table.colActions")}</span>
           </th>
         </tr>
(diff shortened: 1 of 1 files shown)


graphics80:
Pushed `4ac3191f` onto this branch

The new action cell padded with `pl-2`. `audit_i18n.py --strict` fails on physical directional classes and runs as its own CI gate (`web-ci.yaml`, "i18n key audit"), so this branch would have gone red once checks ran. It matters beyond the gate too: `targets.json` lists `ar`, `he`, `fa`, `ur`, and `pl-2` would pad the wrong side of the very column this PR exists to align.

One character: `pl-2` → `ps-2`. `npm run check` and the strict audit both pass on the branch as it now stands.

Found while merging #732, #734, #736 and #738 into a local integration branch to preview them together. One other thing from that: this PR and #732 both rewrite the same action cell, so they conflict. I resolved it locally by keeping this PR's structure and placing #732's `CopyAcceptLinkButton` before the `canMutate` split, since that button is read-only and #732 wants it to survive the archived / can't-author gate. Whichever of the two merges second will hit it for real.

rongxin-liu (approved):
Thanks!

PR #738, opened 2026-08-25
feat(web): flag an out-of-date classroom on the assignments page

### LINKED ISSUE #737: feat(web): the assignments page can't tell you the classroom's data is stale
## Which part would this affect?

`Web app (classroom50.org)`

## Problem or current behavior

The classroom's assignments page renders a freshness line — "Submission data
collected 18 hours ago" — next to **Collect all**. It derives no staleness at
all, so it is permanently in the quiet state no matter how far behind the
snapshot has fallen.

`ClassroomCollectButton` reads only `scores.json` and the last collect run, and
uses them for the timestamp alone. There is no classroom-wide counterpart to
the per-assignment signal the submissions page has had since #364
(`latestAssignmentPush` + `snapshotIsStale` in
`web/src/pages/submissions/dashboard.ts`, derived from the org repo list's
`pushed_at`).

The practical effect: the page whose whole job is "the state of every
assignment at once" is the one page that cannot tell you any of them is out of
date. A teacher scanning the classroom before a grade export sees a timestamp,
reads it as reassurance, and has no way to know a batch of pushes landed after
the last collect — the exact situation the per-assignment badge was built to
surface.

After #733 the two surfaces agree on wording and tone; they still disagree on
behaviour, which is arguably worse than the honest inconsistency before it. The
same "Out of date" badge exists and simply never appears here.

## Proposed solution

Derive classroom-wide staleness and show the same badge.

The comparison must be **per assignment**, not classroom-wide newest-push vs
newest-stamp. The aggregate form produces a false negative in the common case:
hw1 collected a minute ago and hw2 pushed last week but never collected reads
as fresh, because hw1's stamp is newer than hw2's push. One assignment out of
date makes the classroom's data out of date.

Two rules the implementation has to respect, both already established
elsewhere in the codebase:

- **A missing bucket stamp must not borrow a sibling's.** Once any bucket
  carries a `collected_at`, the collector is stamp-aware, so a slug absent from
  the map was genuinely never collected — the precedence `effectiveCollectedAt`
  documents (rule 2). The org-wide run timestamp remains the fallback only for
  a wholly unstamped file.
- **`empty_repo` assignments must be excluded from the question.**
  `collect_scores.py` skips them outright (`if is_empty_repo(entry): continue`),
  so their bucket is never written and never stamped — while their student
  repos exist and carry an accept-time push. Measured like any other assignment
  they would latch the badge on permanently, with no collect able to clear it.
  Their slugs must still feed the sibling-slug guard, or `hw1` starts absorbing
  `hw1-bonus`'s repos. (`no_autograder` is *not* excluded — the collector does
  write its bucket, from detected submissions rather than grades.)

Cost note: the org repo list is a whole-org pagination, but the assignments
table already reads it under the same query key (`AssignmentsTable` calls
`useGetOrgRepos(org)` ungated), so the signal is free on this page.

## Alternatives considered

- **Aggregate max-push vs max-stamp.** Cheaper, and wrong in the case above.
- **Ask the collector for a classroom-level stamp.** Would need a schema change
  and a collector release, and every pre-existing `scores.json` would still
  need the derived path as a fallback.
- **Leave it quiet and rely on the per-assignment badge.** That requires
  opening each assignment to discover which one is behind — the work the
  classroom view exists to avoid.

## Additional context

Follow-up to #733 / #734, which introduced the badge and the shared wording.

Not addressed here, worth its own issue: `AssignmentsTable` runs an
un-memoised `existingGroupRepos` pass per visible row during render, rebuilding
the sibling-prefix list and re-scanning the whole org repo list on every
keystroke in the search box. A shared "repos grouped by slug" memo would serve
that and this signal at once.

## Before submitting

graphics80:
## Summary

The assignments page rendered a freshness line but derived no staleness, so it sat permanently in the quiet state however far behind the classroom's snapshot had fallen. It now shows the same **Out of date** badge the submissions page does.

- **Compared per assignment, not classroom-wide.** New `classroomSnapshotIsStale` asks, for each assignment, whether its repos were pushed after the collect that walked *its own* bucket. The aggregate form (newest push vs newest stamp) would report fresh when hw1 was collected a minute ago and hw2 was pushed last week and never collected — there is a regression test named for exactly that.
- **A missing stamp never borrows a sibling's.** Once any bucket is stamped the collector is stamp-aware, so an absent slug means never collected; the org-wide run timestamp stays the fallback only for a wholly unstamped file. This is the precedence `effectiveCollectedAt` already documents (rule 2), applied per bucket instead of re-derived.
- **`empty_repo` assignments are excluded from the question.** `collect_scores.py` skips them outright, so no stamp can ever arrive, while their student repos exist and carry an accept-time push — measured, the badge would latch on with nothing able to clear it. Their slugs still feed the sibling-slug guard, so `hw1` cannot absorb `hw1-bonus`'s repos. `no_autograder` is deliberately *not* excluded: the collector does write its bucket, from detected submissions.
- **Nothing is claimed while the data is loading.** The badge is gated on the scores read, for the same reason the collected line is suppressed there.
- **One shared strip.** The freshness line and badge move to `components/SubmissionFreshnessLine`, used by both collect surfaces, so the markup can't drift the way it had already started to.

Costs nothing on the wire: `AssignmentsTable` already reads the org repo list under the same query key, so this adds no request. A finished collect now also invalidates that key — `pushed_at` is frozen at page load, so the re-read lets the badge re-derive against the post-sweep repo list. (It cannot see a push the sweep itself missed: the collector stamps after its walk, so a push landing mid-sweep reads as collected until the next push or collect — a collector-side limit, noted in the code.)

Closes #737

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

  - Web: `cd web && npm run check` — passes (4440 unit + 18 browser tests)
  - i18n strict audit — PASS, 0 missing / 0 dead / 0 hardcoded
- [x] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror — n/a. `collect_scores.py` is read here, not changed; the `empty_repo` skip it already implements is what the exclusion mirrors.
- [x] If I added or changed a CLI command or flag, I documented it in the wiki — n/a, no CLI surface changed.

## Review notes

A maintainer-side multi-agent review pass produced seven findings, all addressed in this diff: `classroomSnapshotIsStale` now takes an options object (two positional same-typed slug lists invited a silent transposition); the badge is gated on a failed (non-404) scores read, which previously read as "never collected" and claimed staleness from a read that answered nothing; a just-completed tracked sweep now outranks the lagging completed-run query (mirroring the submissions page), so a legacy unstamped file's badge can't relight right after a successful collect; the post-collect invalidation moved into `hooks/useInvalidateAfterCollect` (the `useQueryClient`/`githubKeys` layer rule); the mid-sweep-push comment no longer overclaims what the invalidation covers; the pre-#694-collector `no_autograder` latch is documented as accepted (the skeleton-drift banner already tells that org to update); and four component tests pin the new behaviors plus the sibling-guard wiring.

Originally stacked on #734; now that it has merged, this branch is rebased onto main and the diff is the staleness derivation alone (one commit).

The rebase surfaced one semantic conflict git couldn't flag: after this branch was cut, #734 gained a test asserting the collect button's tooltip switches to `staleHelp` when stale — coverage for the very branch this PR removes. Here the button stays neutral in every state and the **Out of date** badge carries `staleHelp`, so that test now asserts the constant button tooltip and the badge's help text instead (`DataFreshness.test.tsx`).

This branch went through `/simplify` and `/code-review` before opening. Four findings were applied rather than argued with, and two of them were real defects rather than cleanups — the sibling-stamp borrow and the `empty_repo` latch, both of which had passing tests that never exercised the wiring the app actually uses. The tests added for them fail against the previous behaviour; that was verified by reverting the fix, not assumed.

Deliberately left out, noted in #737: `AssignmentsTable` runs an un-memoised per-row `existingGroupRepos` pass on every keystroke in the search box. It is the hotter of the two O(assignments × repos) paths on this page and wants a shared "repos grouped by slug" memo serving both — but it is pre-existing code and not this diff's to widen into.

### DIFF
--- a/web/src/components/SubmissionFreshnessLine.tsx
+++ b/web/src/components/SubmissionFreshnessLine.tsx
@@ -1,0 +1,66 @@
+import type { ReactNode } from "react"
+import { useTranslation } from "react-i18next"
+
+import { Badge } from "@/components/ui"
+
+// The one freshness strip both collect surfaces render: when the submission
+// data was last collected, an "Out of date" badge when it has fallen behind,
+// and the caller's collect action beside them.
+//
+// Shared because the two surfaces are meant to stay in lockstep — the
+// submissions page's per-assignment DataFreshness and the assignments page's
+// classroom-wide ClassroomCollectButton. They differ only in scope and in
+// which action they hang here, so the wording, the tone, and the layout live
+// once. `stale` is decided by the caller: per assignment on one page, across
+// every assignment in the classroom on the other.
+export type SubmissionFreshnessLineProps = {
+  // Relative "x ago" of the last completed collect for this scope, or null
+  // when nothing has been collected yet.
+  lastCollectedLabel: string | null
+  // The snapshot is (probably) behind — a repo was pushed after the collect
+  // that last walked it.
+  stale: boolean
+  // Drop the collected line while its source is still loading, so it doesn't
+  // flash "not collected yet" before the cached snapshot arrives. The badge
+  // and the action still render.
+  showCollectedLine?: boolean
+  // The collect action, rendered after the badge.
+  children?: ReactNode
+}
+
+export function SubmissionFreshnessLine({
+  lastCollectedLabel,
+  stale,
+  showCollectedLine = true,
+  children,
+}: SubmissionFreshnessLineProps) {
+  const { t } = useTranslation()
+
+  return (
+    <div
+      className="flex flex-wrap items-center gap-x-2 gap-y-1 text-sm text-base-content/70"
+      role="status"
+    >
+      {showCollectedLine && (
+        <span>
+          {lastCollectedLabel
+            ? t("submissions.freshness.collected", { when: lastCollectedLabel })
+            : t("submissions.freshness.neverCollected")}
+        </span>
+      )}
+
+      {/* Staleness is a chip, never the action's tone: collecting is additive
+          and re-runnable, so a danger-toned button would promise a consequence
+          it doesn't have. */}
+      {stale && (
+        <Badge tone="warning" title={t("submissions.freshness.staleHelp")}>
+          {t("submissions.freshness.stale")}
+        </Badge>
+      )}
+
+      {children}
+    </div>
+  )
+}
+
+export default SubmissionFreshnessLine
--- a/web/src/hooks/useInvalidateAfterCollect.ts
+++ b/web/src/hooks/useInvalidateAfterCollect.ts
@@ -1,0 +1,42 @@
+import { useQueryClient } from "@tanstack/react-query"
+import { useEffect } from "react"
+
+import { githubKeys } from "@/github-core/queries"
+import { CONFIG_REPO } from "@/util/configRepo"
+import type { CollectScoresPhase } from "./useTriggerScoreCollection"
+
+// Drops the reads a finished classroom sweep invalidated, so the page
+// re-derives from fresh data: the gradebook (scores.json), the last-run stamp,
+// and the org repo list. Lives here rather than in the page because
+// `useQueryClient`/`githubKeys` stay out of `pages/` (AGENTS.md).
+//
+// `timeout` is only this client giving up on the poll — the run itself usually
+// lands, so refresh there too rather than leaving the page on stale counts.
+//
+// The org repo list is included because `pushed_at` is frozen at page load:
+// without the re-read, a push that landed before the collect's stamp would be
+// invisible and the staleness badge would clear against a list the sweep
+// already left behind.
+const useInvalidateAfterCollect = (
+  org: string,
+  classroom: string,
+  phase: CollectScoresPhase,
+) => {
+  const queryClient = useQueryClient()
+  useEffect(() => {
+    if (phase !== "completed" && phase !== "timeout") return
+    queryClient.invalidateQueries({
+      queryKey: githubKeys.jsonFile(
+        org,
+        CONFIG_REPO,
+        `${classroom}/scores.json`,
+      ),
+    })
+    queryClient.invalidateQueries({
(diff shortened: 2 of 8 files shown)


graphics80:
Held as a draft until #734 merges. It is stacked on that branch — the badge strings (`submissions.freshness.stale` / `staleHelp`) and the shared freshness strip come from there, so on its own the i18n strict audit fails on missing keys.

Once #734 lands I will rebase this onto `main`; the diff then drops the three inherited commits and shrinks to the staleness derivation alone.

rongxin-liu:
@graphics80 Thank you so much for the PR. We'll take it from here unless you have more commits to make. These areas usually touch many surfaces and often bleed into product decisions, so we'll need to evaluate and iterate on them to determine what works best in our view.

graphics80:
Understood — nothing further from me, it is yours to take wherever it needs to go.

I see you already rebased it onto main now that #734 has landed; that was the shrink described in the body, so the diff you are looking at is the staleness derivation on its own.

One piece of context in case it saves a step, since it is the part most likely to bleed into a product decision: `empty_repo` assignments are deliberately excluded from the comparison. `collect_scores.py` skips them outright, so their bucket is never stamped while their student repos still carry an accept-time push — measured like any other assignment, the badge latches on with nothing able to clear it. `no_autograder` is not excluded, since the collector does write its bucket. If those should surface differently, that exclusion is the line to change.

Still marked draft from when it was stacked; that reason is gone, so flip it whenever suits you.

rongxin-liu (approved):

graphics80:
thank you !

PR #933, opened 2026-09-09
fix(gh-student): add timeouts to submit's network calls

### LINKED ISSUE #932: gh student submit hangs indefinitely on a stalled network connection
### Which part is affected?

Student CLI (gh-student)

### What happened?

`gh student submit` can hang indefinitely with no error and no timeout when
a network connection stalls mid-request (dead VPN, silent firewall drop, flaky
wifi). Several of its network/subprocess calls have no deadline at all:

- resolveRepoDefaultBranch(): GET via client.Get(), no context
- fetchRepoPath(): GET via client.Get(), no context; recursive over
  .gitignore/.github, so a deep .github/ tree compounds the exposure
- commitWorkTreeOnRemoteBranch(): `git clone --bare` and `git push` run via
  exec.Command with no context/timeout at all
- pushSubmitTag() / existingSubmitTagAt(): `git push` and `git ls-remote`
  against origin, same issue, runs after the main branch push for any
  tag-mode assignment

Only two calls in the same file (fetchSubmitEntry, resolveAssignmentName)
already use context.WithTimeout  everything else silently blocks forever.
The student sees nothing: no progress, no error, just a stuck terminal until
they Ctrl-C.

Every network/subprocess call submit makes should be bounded by a timeout, so
a stalled connection fails with a clear "network stalled?" error instead of
hanging the terminal indefinitely.

### Steps to reproduce

1. Start `gh student submit` in an assignment repo.
2. While it's mid-flight, stall the connection instead of dropping it
   cleanly.  e.g. block outbound traffic to api.github.com at the firewall,
   or kill a VPN, so the TCP connection stays half-open (accepts, never
   responds) rather than resetting.
3. Observe: the process hangs indefinitely with no output and no timeout,
   whether the stall happens during the default-branch lookup, the
   .gitignore/.github refresh, the clone/push, or (for tag-mode
   assignments) the post-push tag push.

### Version and environment

OS: macOS 26.6.2
gh version: 2.100.0
gh-student version: v1.47.0

### Relevant logs or output

```shell
none. It is a bug. no error, no timeout, just a stuck process.
```

### Before submitting

aandhare:
## Summary

`gh student submit` could hang indefinitely on a stalled network connection (dead VPN, silent firewall drop): the default-branch lookup, the recursive teacher `.gitignore`/`.github` refresh, the bare clone + push, and the tag-mode `ls-remote` + tag push all ran with no deadline. Every one of them is now bounded.

- REST calls (`resolveRepoDefaultBranch`, `fetchRepoPath`) go through one `getBounded` helper using `RequestWithContext` with a per-request deadline, so a large `.github/` tree is never penalized for its size.
- Git network calls use git's own stall detector (`GIT_HTTP_LOW_SPEED_LIMIT=1`, `GIT_HTTP_LOW_SPEED_TIME=30`): a dead HTTPS connection fails in about 30 seconds while a slow but progressing clone or push is left alone. A generous 10-minute ceiling remains as the backstop for SSH remotes, and `pushSubmitTag` has its own 30-second bound.
- All git invocations share one `gitCmd` constructor (`exec.CommandContext` + `cmd.WaitDelay`, so an orphaned `git-remote-https` can't hold `Wait()` open after git is killed) and one `gitErr` wrapper. A deadline surfaces as a student-readable error that says what to do next; Ctrl-C (which cancels the root context) is not reported as a network problem.
- Timeouts live in a single `var` block; the ones tests shrink are vars, the existing ones stay consts.

Tests cover each stalled path with a server that accepts but never answers, including separate proofs that the stall detector fires before the ceiling and that the ceiling fires with the guidance message.

Closes #932

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

- [ ] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass.
- [ ] If I added or changed a CLI command or flag, I documented it in the wiki (not in a README).
- [x] My commits follow Conventional Commits (e.g., `feat(web): ...`, `fix(gh-teacher): ...`).

### DIFF
--- a/cli/gh-student/internal/submitcmd/orchestration_test.go
+++ b/cli/gh-student/internal/submitcmd/orchestration_test.go
@@ -1,6 +1,7 @@
 package submitcmd
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"net/http"
@@ -48,7 +49,7 @@ func TestFetchRepoPath(t *testing.T) {
 	client := githubtest.NewTestClient(t, server)
 
 	dst := t.TempDir()
-	if err := fetchRepoPath(client, dst, "o", "r", "main", ".github"); err != nil {
+	if err := fetchRepoPath(context.Background(), client, dst, "o", "r", "main", ".github"); err != nil {
 		t.Fatalf("fetchRepoPath: %v", err)
 	}
 
@@ -78,7 +79,7 @@ func TestFetchRepoPath_RejectsNonBase64(t *testing.T) {
 	t.Cleanup(server.Close)
 	client := githubtest.NewTestClient(t, server)
 
-	err := fetchRepoPath(client, t.TempDir(), "o", "r", "main", "file.txt")
+	err := fetchRepoPath(context.Background(), client, t.TempDir(), "o", "r", "main", "file.txt")
 	if err == nil || !strings.Contains(err.Error(), "unsupported encoding") {
 		t.Fatalf("err = %v, want an 'unsupported encoding' error", err)
 	}
--- a/cli/gh-student/internal/submitcmd/submit.go
+++ b/cli/gh-student/internal/submitcmd/submit.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"net/http"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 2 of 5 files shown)


rongxin-liu (approved):
Thanks for the contribution.

PR #961, opened 2026-09-11
fix(cli): print concise usage errors and one-line flag help

### LINKED ISSUE #948: `gh-teacher` flag help text too verbose

Teacher CLI (gh-teacher)

If execute `gh-teacher assignment add`:

Image

the flags take most of the page, so the good designed part (usage, examples pattern, which follows https://bettercli.org/design/cli-help-page/) in the beginning is not shown in a medium terminal window.

### Proposed solution

The flag documentation is probably autogenerated, but every text could be shortened to a one-liner.

### Alternatives considered

_No response_

### Additional context

_No response_

### Before submitting

aandhare:
## Summary

Fixes the `--help` and usage-error output of both CLIs (issue #948), starting from @aandhare's diagnosis and templates and extending it to the copy, the shared module, and the tests.

### Behavior

- **Usage errors are concise.** A wrong arg count or unknown flag prints the usage line and `Run '... --help' for details and examples.` instead of every flag. `gh teacher assignment add` with no arguments prints 6 lines; it printed 46.
- **A mistyped subcommand fails.** `gh teacher assignment ad ...` used to print the group's full help and exit 0, because Cobra never validates args on a non-runnable command. Groups now name the unknown command, suggest the closest match, print the concise usage, and exit 1. `gh teacher assignment help` (the word, not the flag) takes the same path and exits 1, as `gh pr help` does.
- **Flag descriptions wrap** to the width of the terminal the help is written to, capped at 100 columns; help sent to a pipe or a file wraps at 100. pflag's default never wraps, so a long description ran off the screen as one line.
- **Every flag description is one line.** All 152 flags across both CLIs follow the gh CLI's shape, which the writing style guide already asks for: a capitalized fragment, no period, enums as `{a|b|c}`, at most 100 characters. Several descriptions used backticks for inline code; pflag treats the first backtick pair as the flag's type placeholder, so `--feedback-pr` was rendering as `--feedback-pr gh teacher init` and skewing the whole column.
- **`assignment add` Long is shorter.** It keeps the constraints a teacher must know first (slug rule, what is required, re-add semantics, hide-by-default and `--locked`, the three grading modes as a list) and the safety caveats (public repos expose student work and are not autograded, a Pages site is public even on a private repo, admin permission, every flag `--empty-repo` excludes), and links to the wiki page that already documents every flag.

`gh teacher assignment add --help` at 100 columns: 208 visual lines on `main`, 109 here, with Usage at line 47 and Examples at 50.

### Implementation

- `cli/shared/ghhelp` owns the templates and the group guard, so `gh-student` gets the same behavior; each CLI's `newRootCmd` calls `ghhelp.Install(root)`. The help template is derived at runtime from Cobra's defaults (`UsageTemplate()` and `HelpTemplate()` on a bare command) with the flag sections swapped for the wrapping variant, rather than a frozen copy; the concise usage-error template is a short hand-written one. Help keeps the gh CLI's order: description, then usage, then flags.
- The group guard works by giving each group a `RunE`, which makes it runnable in Cobra's eyes. Both templates guard the usage line with `and .Runnable (not .HasAvailableSubCommands)` so a group keeps its single `group [command]` line instead of gaining a `group [flags]` line.
- Wrap width comes from the writer that receives the help (`cmd.OutOrStdout()`), read through a small seam tests can pin, so `--help` into a buffer renders the same on every machine.
- `ghhelp.Lint` walks a command tree and reports copy violations: description over 100 characters, lowercase start, trailing period, em dash, `...`, banned words, or a decorative backtick pair.

### Tests

- `TestHelpCopyConventions` in each CLI runs `ghhelp.Lint` over the real tree.
- `TestUsageErrorIsConciseOnRealTree`, `TestGroupTypoFailsOnRealTree`, `TestGroupHelpOnRealTree`, and `TestHelpRendersWrappedFlagsOnRealTree` execute the real root into a buffer, so a dropped `Install` call fails the suite. They assert the usage error is an error, the group typo fails with a suggestion (and exit code 1 through `cliutil.ExitCodeFor` in gh-teacher), group help has no `[flags]` line, every rendered line fits the 100-column cap, and the safety caveats still appear in `assignment add --help`.
- `cli/shared/ghhelp` covers template derivation (fails if Cobra's default template changes shape, including the usage-line guard), the concise usage path, wrapping at the cap and at a pinned 72 columns, the group typo, bare-group, and group-help cases, and every lint rule.

### Considered and not done

The git-style `[--flag] [--flag] ...` usage line quoted in the issue. With 19 flags it would wrap to a dozen lines and bury the positional arguments; Cobra's `[flags]` placeholder plus the one-line Flags block is the better fit for commands this size.

Closes #948

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Refactor / maintenance

## Checklist

- [x] I built, tested, and linted the module(s) I touched: `go build ./... && go test ./... && golangci-lint run` and `golangci-lint fmt --diff` are clean in `cli/gh-teacher`, `cli/gh-student`, and `cli/shared`.
- [ ] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror, and the parity tests pass. (No contract change.)
- [x] If I added or changed a CLI command or flag, I documented it in the wiki. (No new commands or flags. The wiki's `gh-teacher` page is a hand-maintained reference that already documents every flag; the shortened `assignment add` Long now links to it.)


rongxin-liu (approved):
Thanks for working on this. I made a few changes to fully address the issue.

PR #971, opened 2026-09-12
feat(gh-teacher): re-issue an expired invitation and drop a dead pending row

### LINKED ISSUE #970: roster invite and cancel-invite deadlock after an org invitation expires
GitHub organization invitations expire after 7 days. If one lapses before the student accepts, there is no CLI path to re-issue it.

**State After Expiry**

- GitHub has no pending invitation. `GET /orgs/{org}/invitations` no longer lists it.
- `roster.csv` still holds the student's pending row (email present, `username` empty).
- A leftover `invite-<hash>` metadata team remains.

**Behaviour**

- `gh teacher roster invite <org> <classroom> <email>` refuses with "... is already invited to ... and nothing was sent", because the pending roster row reads as an outstanding invitation.
- `gh teacher roster cancel-invite <org> <classroom> <email>` refuses with "no pending invitation for ..., nothing was cancelled", because it only acts on a live invitation.
- `gh teacher roster sync <org> <classroom> --write` does delete the orphaned `invite-<hash>` metadata team, but it leaves the pending roster row, and `roster invite` still refuses afterwards.

Each command defers to the other's precondition, so neither can act.

**Workaround**

Issue the invitation directly, attaching the classroom team so that accepting still enrols in one step:

```
gh api -X POST /orgs/<org>/invitations \
  -f email=<email> -f role=direct_member -F 'team_ids[]=<classroom team id>'
```

The next `roster sync` then fills the pending row.

**Possible Fixes**

- Have `roster invite` notice that the roster row is pending while GitHub has no matching invitation, and re-issue rather than refuse, perhaps behind a `--resend` flag.
- Or let `cancel-invite` clear a stale pending row when there is no live invitation, so that `roster invite` can proceed.

Observed on gh-teacher v1.44.0. I have not checked whether v1.48.x still behaves this way.

aandhare:
## Summary

GitHub drops an organization invitation after 7 days. The roster row it left behind survives, but `roster invite` refused the address because the row looked pending, and `cancel-invite` refused because there was no live invitation to revoke, so the CLI had no way to re-invite the student.

`roster invite` (single address and `--file`) now judges a pending row by asking GitHub instead of by the row's shape:

- The invitation is still on the org's pending list: refuse as before. The list is org-wide, so the message says the invitation may be another classroom's, and names `roster sync --write` and `cancel-invite` as the ways forward.
- The invitation is gone but the per-invite metadata team holds a valid record and a member: the student accepted and only a sync is missing, so refuse and point at `roster sync --write`. Sending again would trip the team's not-empty check with a message telling the teacher to remove the invitee, which would destroy the only email-to-account mapping.
- The team is gone, empty, or still carries the provisional description an interrupted run leaves behind (with the teacher GitHub added on create as its only member): nothing backs the row, so send a new invitation against the same row. The metadata team is adopted and healed if it survived or recreated if the sync already collected it, no second row is written, and GitHub's expired record for the address is dismissed once the send is confirmed (or once GitHub answers that the address is already covered). This is the web app's Re-invite.
- The team's record was edited by hand and it holds a member: that member may be the student, so the send is refused for the teacher to check, the same anomaly `roster sync` reports.

GitHub uses one 422 for three answers on the invitation POST. `InviteOrgByEmail` now reads the body the way the web's `classifyInvitation422` does: "already" is a skip, the organization's invitation limit is treated like a rate limit (the fresh team is kept for a retry, nothing is dismissed, bulk defers the rest and exits 2), and any other validation failure is reported with GitHub's reason instead of being read as "already invited".

A 403-shaped secondary rate limit used to be stripped by the membership classifiers before the invite path could see it, so a throttled read reported an admin-access problem and a bulk run counted the row as failed. Both classifiers now leave a rate limit wrapped so `IsRateLimited` still recognizes it.

Both GitHub lists are read lazily, so an address with no pending row touches neither and gains no new way to fail. The failed-invitation list is owner-only: a plain 403 or a 404 reads as nothing to dismiss, a rate-limited 403 or any other read or DELETE failure is a warning, and none of it can fail a send that already went out.

`roster sync` also dismisses expired records for the addresses it recovers as accepted. A dry run reports them and counts them as pending work (exit 2), and `--write` dismisses them after the roster commit lands.

`roster remove <org> <classroom> <email>` drops the pending row a lapsed invitation left behind, the web app's Remove row. It reuses the same classification: a live invitation is refused and left for `cancel-invite` (the message notes it may be another classroom's, since the pending list is org-wide), an accepted-but-unsynced student is refused and left for `roster sync --write`, a hand-edited team with a member is refused for a human, and only a dead row is dropped (every duplicate pending row for the address, in one rebase-retried commit). The leftover metadata team is then deleted, but only on a fresh check taken after the commit, as the sync's teardown does: a web Re-invite or an acceptance in that window keeps the team, since it now maps a live invitation. GitHub's expired record is dismissed last. Both teardown steps warn rather than fail. The address is canonicalized before auth like the invite paths, so a form GitHub rejects fails up front instead of matching no row; an address that sits on a row with an account is refused with the username form to run instead; an absent row is a no-op that reads neither GitHub list. This is the issue's second suggested fix, and without it the only way to drop such a row was to edit `roster.csv` by hand.

`cancel-invite` still acts only on a live invitation. Its no-pending hint lists the three ways out (sync if they accepted, `roster invite` if it expired, `roster remove <email>` to drop the row). Every "record them" hint across the commands now names `roster sync --write`, since a dry run records nothing.

New `membership.ListFailedOrgInvitations` reads `GET /orgs/{org}/failed_invitations`; dismissal reuses the existing invitation DELETE. The CLI teacher guide, command reference, and troubleshooting pages are updated to match.

Closes #970

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

- [ ] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass.

### DIFF
--- a/cli/gh-teacher/internal/membership/membership.go
+++ b/cli/gh-teacher/internal/membership/membership.go
@@ -139,6 +139,32 @@ func ListPendingOrgInvitations(client githubapi.Client, org string) ([]PendingOr
 		func(path string, err error) error { return ClassifyMembershipReadError(path, subject, err) })
 }
 
+// FailedOrgInvitation is one GET /orgs/{org}/failed_invitations element: an
+// invitation nobody accepted within GitHub's 7 days, or one GitHub couldn't
+// deliver. Keyed by login or email exactly like a pending one.
+type FailedOrgInvitation struct {
+	ID           int64  `json:"id"`
+	Login        string `json:"login"`
+	Email        string `json:"email"`
+	FailedReason string `json:"failed_reason"`
+}
+
+// ListFailedOrgInvitations walks the org's failed-invitation list, the record
+// GitHub keeps of an expired invitation once it leaves the pending list. Owner
+// only. Dismissing a record is CancelOrgInvitation on its id (the same DELETE;
+// GitHub's UI calls it "dismiss"). The raw HTTP error is preserved (not
+// classified) so a caller can read a 403/404 as "no list to sweep": these
+// records are bookkeeping, and whether a failed read matters is the caller's
+// call.
+func ListFailedOrgInvitations(client githubapi.Client, org string) ([]FailedOrgInvitation, error) {
+	base := fmt.Sprintf("orgs/%s/failed_invitations", url.PathEscape(org))
+	return githubapi.PaginateAll[FailedOrgInvitation](client, githubapi.ListPerPage, githubapi.ListMaxPages,
+		func(page int) string {
+			return fmt.Sprintf("%s?per_page=%d&page=%d", base, githubapi.ListPerPage, page)
+		},
+		func(path string, err error) error { return fmt.Errorf("GET %s: %w", path, err) })
+}
+
 // InvitationTeamRef is one element of GET /orgs/{org}/invitations/{id}/teams:
 // the teams an invitation will add its invitee to on acceptance.
 type InvitationTeamRef struct {
--- a/cli/gh-teacher/internal/membership/membership.go
+++ b/cli/gh-teacher/internal/membership/membership.go
@@ -19,6 +19,8 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"regexp"
+	"strings"
 
 	"github.com/foundation50/gh-teacher/internal/cliutil"
 	"github.com/foundation50/gh-teacher/internal/githubapi"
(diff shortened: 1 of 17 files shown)


rongxin-liu (approved):
Thanks for the PR. I made some changes.

PR #981, opened 2026-09-14
fix(gh-teacher): show a failing run test's stdout and stderr together

### LINKED ISSUE #910: Show output for failed tests

Autograder / grading

It seems currently, when test fails, it reports only stderr, but not stdout, at least for declarative Run command grader.
This is not convenient when running test involves additional logic, like producing its own more complicated diff, which usually goes to stdout and not stderr.

### Proposed solution

Print both stderr and stdout on failure.
Ideally, they should be captured together and not as separate streams, to preserve the same meaning in case of interleaved stdout and stderr messages.

### Alternatives considered

In my case I had my own diff logic after running `clang-format`, which in my `Makefile` was sent to stdout.
Either changing Run command from `make` to `make 2>&1` or forwarding output of `diff` directly to stderr do workaround the problem.

### Additional context

_No response_

### Before submitting

aandhare:
## Summary

A failing `run` test showed only stderr and dropped stdout, so a command that prints its result to stdout (a diff, a self-check) showed nothing useful on failure.

Both streams now show, captured as one in the order the command wrote them, the way a terminal or a CI log shows them. This is the model GitHub Actions run steps and GitHub Classroom's own command grader use: a `run` test is a command judged on its exit code, so its full output, in order, is the signal. The same applies to `python` tests and to every `setup` command. `io` tests keep the two streams apart because the comparison reads stdout on its own. Grading is unchanged.

Ordering is bounded by the program's own buffering, the same as `2>&1`: a program that block-buffers stdout when piped may show its errors before its earlier output. Forcing terminal ordering would need a PTY, which changes how programs behave, so the wiki documents the caveat instead. The Release body and Summary keep the first 2,000 characters of the combined stream, the same per-block clip as before, so a verbose build can lose its trailing error line there; the wiki says so next to the merge note and points to the 100,000-character grade log. No new configuration is needed and no contract key is added.

One function (`compose_detail`) feeds every surface, so this covers the release body, the Actions log, and the run summary; the web app links to the release page and needs no change.

Closes #910

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

- [ ] If I changed a cross-binary contract, I updated `schemas/*.schema.json` **and** every mirror (Go / Python / TypeScript), and the parity tests pass. (No contract change; only two schema descriptions were reworded.)
- [ ] If I added or changed a CLI command or flag, I documented it in the [wiki](https://github.com/foundation50/classroom50/wiki) (not in a README). (No CLI change; the Autograding Basics and FAQ pages were updated.)


rongxin-liu (approved):
Thanks!

PR #982, opened 2026-09-14
feat(gh-teacher): add combine-output to interleave a run test's streams

aandhare:
## Summary
Adds an opt-in `combine-output` per-test flag (run tests only) that merges stderr into stdout (like 2>&1), so a failing run test's output shows one interleaved stream in the order it was printed rather than separate blocks. Off by default; io and python reject it (io grades by comparing a clean stdout). Interleaving follow-up to #910.

This builds on #981 (the show-both-streams fix) and currently includes its commit, so the diff reads cleanest once #981 has merged. I'm happy to rebase this onto main to drop that commit once #981 lands, or whenever suits your review.

Scope (a cross-tool contract change, so all mirrors move together): schema (tests-v1 + assignments-v1 in lockstep), Go validator plus the `--combine-output` flag on `assignment test add`, the runner (merge + render), the web form (a run-only control) and its round-trip, docs, and tests in all three.

### Why not force perfect ordering
Ordering for a command that block-buffers its own stdout is bounded by that buffering, the same as any 2>&1. The only way to force it is a PTY, which makes the command think it is interactive and changes its output, so we don't. Grading is unaffected: run tests score on exit code.

## Type of change
- [x] New feature

## Checklist
- [x] Built, tested, and linted (Go, Python autograders/skeleton, web typecheck/vitest/prettier/eslint)
- [x] Cross-binary contract updated in lockstep (schema + Go + Python + web), parity test green
- [x] Documented in the wiki
- [x] Conventional Commit


rongxin-liu:
Thanks for working through the interleaving question here. After looking at how GitHub Actions run steps and GitHub Classroom's command grader present a command's output (one stream, as the terminal shows it), we went with interleaving by default for `run` and `python` tests and for `setup`, with `io` tests keeping separate streams because the comparison reads stdout. That landed in #981, so a `combine-output` flag is no longer needed and there is no new key for every reader to mirror. Closing this one in favor of #981; your buffering analysis in the description was right and is now reflected in the wiki.

PR #988, opened 2026-09-14
fix(web): open dropdown triggers in Safari

### LINKED ISSUE #987: Actions dropdown does not open when clicked in Safari
### Which part is affected?

Web app (classroom50.org)

### What happened?

On an assignment’s Submissions page, clicking **Actions** in Safari does not open the dropdown menu.

The same menu opens in a Chromium-based browser.

Clicking **Actions** opens the dropdown.

### Steps to reproduce

1. Open an assignment’s Submissions page in Classroom 50 as a teacher.
2. Click **Actions** in Safari.
3. Observe that the dropdown does not appear.

### Version and environment

- macOS
- Safari 26.6.2
- Affected page: assignment Submissions page on classroom50.org

### Relevant logs or output

```shell
Reproduced in an authentication-free local fixture using Classroom 50's
actual Button, DropdownMenu, and application CSS.

Source commit: 49372ac46ac8b0c8a8ffddb8dbcbcb8ad1769445
daisyUI: 5.7.32
Safari: 26.6.2

Observed in real Safari:
- Original trigger: clicking leaves document.activeElement as BODY.
  The menu remains hidden.
- With onClick={(event) => event.currentTarget.focus()}:
  document.activeElement becomes BUTTON and the menu opens.

Playwright results (Chromium and WebKit):
                     Original    Explicit-focus candidate
Chromium             3 passed    3 passed
WebKit               3 failed    3 passed

The tests check opening, selecting an item, and closing via an outside
click followed by reopening. All original WebKit failures occur because
the menu does not become visible after clicking Actions.

The fixture uses the shared production components, but does not render
the full SubmissionsActionsMenu component or authenticated page.

daisyUI documents the Safari button-focus caveat:
https://daisyui.com/components/dropdown/?lang=en

Relevant source:
https://github.com/foundation50/classroom50/blob/49372ac46ac8b0c8a8ffddb8dbcbcb8ad1769445/web/src/pages/submissions/SubmissionsActionsMenu.tsx
```

### Before submitting

hisazumi:
## Summary

In Safari, clicking **Actions** on the Submissions page did nothing. daisyUI dropdowns open on `:focus-within`, and Safari does not focus a `<button>` on click, so the menu never opened.

The fix gives the trigger an explicit `tabIndex={0}`. WebKit's Cocoa port makes a form control mouse-focusable only when tabindex is set explicitly (`HTMLFormControlElement::isMouseFocusable`), and this is the recipe `SplitButton` and `LoginLanguageMenu` already use. Because focus now happens natively on mousedown, it does not depend on the shared `Button`'s loading guard. The bulk-selection **Actions** trigger (`BulkSelectionCluster`) had the same defect and gets the same attribute. Chromium and Firefox already focus buttons on click, so the attribute is a no-op there.

Closes #987

### What changed from the first revision

The original commit (kept intact for credit) focused the trigger from an `onClick` handler and added a WebKit vitest project plus WebKit installation in the CI browser lane. The follow-up commit on top replaces that with the tabindex fix and removes the WebKit lane: Playwright's Linux WebKit is the GTK/WPE port, which focuses buttons on click natively, so that lane passed with or without the fix. The Chromium browser test from the first revision stays as regression coverage for the menu's pointer interaction, and a happy-dom test now guards the tabindex on the trigger.

Verified in real Safari, this does not run as a CI test: only the macOS WebKit port reproduces the bug.

### Deferred to a separate PR

- A shared dropdown-trigger primitive (or `Button` prop) so the Safari attribute lives in one place instead of at each call site.
- Guarding any future extra vitest browser project in `browserProjectWiring.test.ts`.
- Keyboard-open coverage for the menu in a real browser.

### Validation

- `cd web && npm run check`: passed (437 test files, 5758 tests; tsc, eslint, prettier, arch:validate).
- `npx vitest run --project browser src/test/submissionsActionsMenu.browser.test.tsx`: 3 passed in Chromium.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / maintenance

## Checklist

- [x] I built, tested, and linted the module(s) I touched: `cd web && npm run check`
- [ ] Cross-binary contracts: not changed.
- [ ] CLI commands or flags: not changed.


rongxin-liu (approved):
Thanks for your contribution. I’ve made a few changes.

PR #1021, opened 2026-09-17
fix(web): section-filter unsubmitted group and team rows

### LINKED ISSUE #1020: Section filter shows teams with no members in the selected section
### Which part is affected?

Web app (classroom50.org)

### What happened?

On a group assignment's submissions page, filtering by a section also shows
teams that have no members in that section. The students who are actually in
the selected section show correctly, but unrelated teams from other sections
appear alongside them.

A team shows under a section only if at least one of its members is in that
section, the same rule submitted rows already use.

### Steps to reproduce

1. Open a group or team assignment's submissions page.
2. Have some teams that formed but have not submitted, with members outside
   the section you will filter by.
3. Set the section filter to that section.
4. Teams with no member in that section still appear in the list.

### Version and environment

Web app (classroom50.org), current release. Not browser specific (client-side
filtering).

### Before submitting

- Searched existing issues, not a duplicate.
- No tokens, secrets, or student data included.

aandhare:
Fixes #1020

## What

On the submissions page, the section filter was applied to submitted rows and individual non-submitters, but not to unsubmitted group repos or teams with no repo yet. Those two were filtered only by the search box, so a formed team showed under any section, even one with no member in it.

## Change

- `filterDisplayList` now section-matches the group repo and team-without-repo lists through a `groupMembersOf` resolver, using the same "any member in the section" rule `rowInSection` already applies to submitted rows.
- `SubmissionsPage` supplies the resolver in both modes. Team mode uses live team membership by `group-<n>`. Legacy group mode uses the founder plus the repo's collaborators, which `useGroupRepoMemberLogins` now also returns per repo (it already fetched them to build the non-submitter exclusion set). No new GitHub reads.
- Unknown membership keeps a group visible, so a team or legacy group whose members are still loading (or whose collaborator read failed) does not flash out of the filtered view.
- Array identity is preserved when no section or search filter is active, so callers that memoize on the lists are not re-triggered.

## Tests

- Unsubmitted legacy group repos are filtered by any member's section: a group founded by a B2 student with a B1 collaborator shows under B1 and B2.
- Teams are filtered by member section with the matching member listed second, so a first-member-only implementation fails; a team whose members have not loaded stays visible.
- A team whose members loaded as empty is hidden, unlike unknown membership.
- Section and search compose on groups; section `all` ignores membership.
- `useGroupRepoMemberLogins` returns lowercased members per repo and omits a repo whose read failed.


rongxin-liu (approved):
Thanks!

